PR #6, opened 2015-02-11
Add CORS support to Dicoogle web services

Enet4:
This adds support to Cross-Origin Resource Sharing, so that web browser applications may perform cross-origin requests to these services.


bastiao:
@tmgodinho since we will deprecate the RESTlet web services, should stop allowing modifications on that? 
@fredericojssilva what are the status of web services implementation in Jetty?

Enet4:
I honestly didn't know they were going to become deprecated, but it appears to be what we currently have. In addition, I had to add CORS support for the JavaScript web service access wrapper.

Enet4:
I will close this PR for now, as it could bring some safety issues. Hopefully, this will be addressed in the new implementation.

bastiao:
Can you fix it in the Jetty? 
Instead of allow it from all, we can allow for a specific domain.

Enet4:
I haven't looked into the new implementations at all so far. But yes, specifying accepted cross-origin domains would become a requirement.

PR #7, opened 2015-02-14
Feature/api

fredericojssilva:
Falta os serviços de start & stop de serviços e plugins.
Falta tambem rever a compliance com QIDO.RS


Enet4:
I have seen the new code using custom HTTP status messages, which is a bad practice. Do you have the full API described somewhere? Perhaps writing it in the wiki would be quite useful.

bastiao:
Indeed, it is a good idea.
Check it out and feel free to give your contributions: 
https://docs.google.com/document/d/1w9OP5zct-K7Ln1zsvvrEf0xfEkYLPyTQ_ihiq4O13qA/edit?usp=sharing

bastiao:
@fredericojssilva can you finish the QIDO-RS, and the missing services of the plugin? To merge the pull request?

Enet4:
One of my concerns is how errors are being handled right now. If the underlying format standard does not contemplate such cases, then we should at the very least have a dedicated data type like a `{code,message}` tuple.

Enet4:
@fredericojssilva I've started a [wiki page](https://github.com/bioinformatics-ua/dicoogle/wiki/Dicoogle-Web-Service-API) for the API.

bastiao:
Folks, it is better to merge this branch. 
It will be easier for other branches pick the changes directly from dev. What do you think?

bastiao:
Following our F2B discussion, we decided to merge.

PR #9, opened 2015-02-24
Add missing ladda and jquery-quicksearch scripts

Enet4:
A few JavaScript libraries for the web UI were missing. In Ladda, this might have happened because our base .gitignore ignores all folders named "dist". I am not quite sure about what happened to QuickSearch. Nevertheless, this PR includes both.


bastiao:
It happen due to the migration between private and public repository.

bastiao:
Nice catch. Your branch was not synchronised with dev, so I updated.

PR #10, opened 2015-02-24
Make username and search text inputs autofocused

Enet4:
Simply places `autofocus` in reasonable input elements.


bastiao:
Nice! Merged with dev and merged.

Issue #20, opened 2015-03-08
Web UI Plugin Interface


Enet4:
With the deprecation of the Swing graphical plugin interface and the uprising of future use cases for Dicoogle, there comes a need for extendability at the web user interface level. Web Service plugins can provide web interfaces, but these are not integrated into the rest of the Dicoogle web application.

Here are some of the features that we are trying to cover:

- A user wants to perform a query by example (using content-based image retrieval). The best step into performing the query would be dragging and dropping an image into a file upload element in the Search page. The current querying interface is fixed to a basic querying method and an advanced querying method, which lack a means to send images. Future querying engines should be able to expose other query parameters in the interface itself.
- There may be cases where the querying engine(s) in use affect(s) what contents lie in a result. For instance, a CBIR query operation will provide results with a _similarity_ property, and may not provide the remaining information about the results (such as patient name, patient/study/series ID). On the other hand, there are other interesting ways to show results that might not be covered in the web site, such as showing a grid instead of a list.

Below is a diagram with my proposal. The Dicoogle web page will provide specific UI slots for plugins to be inserted into (multiple plugins in the same slot could be separated with tabs). A Web UI plugin would compose a JavaScript module and a package descriptor (using a "package.json" typically seen in Node.js). An example of a package descriptor can be seen below. At the `dicoogle` subtree, it must contain at least the file name of the module and a pointer to the slot meant to be attached to.

```json
{
  "name" : "cbir-query",
  "version" : "0.0.1",
  "description" : "CBIR Query-By-Example plugin",
  "dicoogle" : {
    "caption" : "Query by Example",
    "slot-id" : "query",
    "module-file" : "module.js"
  }
}
```

At the level of the actual module, a `DicoogleWeb` object would be exposed with some operations only relevant to the web interface. `issueQuery` would perform the query (thus passing all the data needed by the engine) and trigger a result plugin by invoking an event in the result slot. The `onResult` event function, implemented by each result web UI plugin, should take a list of results and expose them freely.

![webui](https://cloud.githubusercontent.com/assets/4738426/6602028/0edf7fbe-c811-11e4-8e44-f3db0fb87b80.png)

Let this issue be used to design and discuss Dicoogle's Web UI plugin interface. I will add more details as more suggestions and ideas come up.

bastiao:
Good idea.

According to our discussion I'm suggestion to have kind of Javascript Plugin that should be loaded  at the bootstrap stage. 

[...]

PR #30, opened 2015-04-03
Web UI Plugin support - Dicoogle Web Core + test plugins

Enet4:
This PR introduces the architecture for an extendable web UI interface (#20).

- Added web plugin management to Dicoogle.
- Added a web service for retrieving the web UI plugins.
- Added Dicoogle Web Core, a JavaScript module for loading web UI plugins and exposing them an API. 
- Some sample plugins are provided (they must be installed locally beforehand, by passing the full directory to the WebPlugins base directory.


bastiao:
Very nice work. I really believe that it will be really useful in Dicoogle future. 
Comments: 
```
return "{\"plugins\":[" + String.join(",", pkgList) + "]}";
```

Probably it can work (you need to check): 
```
return "{\"plugins\":[" + StringUtils.join( pkgList, ",") + "]}";
```

Please, use compatible syntax with JDK 7, since it lives in Dicoogle. 

Minor issues that does not blocks the merge and can be fixed in a subsequent PR, but if you reply now, it will be nice:

- why are you tracking filesizes.txt? We really need it?
- in package.json of web core why "dep/jquery.js”? are you using dep path?
- using <dicoogle-slot> seems nice. but why use a new tag instead of a single div (compatible with all browsers) and use a class or id? what are the real advantages?
- since you are using AMD, bower could be nice to fetch the dependencies dynamically  (for instance, in dicoogle-webcore/test)? 
- Putting some information to test the demo/test in the dicoogle-webcore/test will be nice. You guide the developers to do one new module, but it still confusing to learn, how to run and test it.

Enet4:
Yeah, thanks for the pointers.

- Yeah, that isn't needed. I will get rid of it in the next commit.
- That used to be a dependency, but not anymore. I've scratched that off for the next commit.
- The use of custom elements is compatible in HTML5, and can bring some potential in the long run. For instance, we can easily detect whether an element with this tag was attached. 
- I don't use bower. Does that run in the browser? Otherwise, what exactly are you suggesting? All the matter of dynamic module loading and dependency management has been a concern to me, and a tough one to get it right.
- The goal of the test chamber was to do just that. I'll see what I can do.

bastiao:
Answer to point 3: Ok. 
Answer to point 4: dependency management to dicoogle-webcore/test. Not seeing the point of have all dependencies there (you can use the npm as well I guess).
Answer to point 5: my point was only documentation to run it or have a kind of dicoogle-webcore/demo if you prefer.  

Cheers!

Enet4:
Sure thing, I can install dependencies with npm. But it would have been ideal to just let the plugins state what dependencies to load.

bastiao:
Yes, make sense!

bastiao:
Is it ready to merge?

PR #32, opened 2015-04-05
Add JavaScript Dicoogle service client

Enet4:
This is a web service client API to Dicoogle, for use in JavaScript applications. Both browser JavaScript and Node.js are supported.


bastiao:
It will be moved to another repository and sub modules will be added according to a F2F discussion with @Enet4

PR #33, opened 2015-04-07
Some cleanup

Enet4:
- Marked rGUI related content as deprecated. They still cannot be removed with ease due to a bad dependency between some business logic (e.g authentication) and the rGUI. (to be addressed with #34)
- Removed an unused `SearchServlet` (the old one, wasn't doing its job anyway).
- Moved some classes unrelated to the remote GUI out of the rGUI package.


bastiao:
Nice!

Issue #51, opened 2015-05-22
[webapp] RunningIndexTasks not handling tasks properly


Enet4:
The current implementation of RunningIndexingTasks seems to work ok with our Lucene plugin. However, it reveals some flaws when applied to other plugins or simply more than one:

- When calling `index(URI)`, each indexing provider is given a task of its own. However, these are not assigned a distinct UUID.
- It is currently not possible to know whether indexing tasks that do not give a task progress have finished. It is expected that their progress value is always -1. The servlet must provide a way to know when such tasks are finished.

PR #52, opened 2015-05-22
Fix #49 and #51; Also force indexing no longer waits

### LINKED ISSUE #49: [webapp] ForceIndexing servlet is assuming an IndexReport from index()
The new servlet for forcing the indexing procedure is assuming that all index operations will return an `IndexReport`. This is currently not guaranteed by the SDK, and for now we must admit that a plain `Report` might be returned by the indexing plugin.

I have also seen the use of Java reflection to iterate through the report's getters and invoke them to obtain a complete list of entries from the report. Although I understand the concern of supporting a variable number of properties, using Java reflection here is not a good practise: it might retrieve information unrelated to the report (the plugin developer is currently oblivious to Dicoogle's core using reflection in such a manner), and may slightly cripple performance. Later on, version 3.0 will hopefully address this class hierarchy with a redesign of the SDK.

I suggest using a plain `instanceof IndexReport` for now to identify key cases where an IndexReport is obtained, and interpret a bare `Report` as "finished indexing".

By the way, as mentioned some time ago, the ForceIndexing servlet should not wait for completion.

Enet4:
The title says it all. All index-related operations were tested with 2 different index plugins, one of them not providing a task progress (CBIR).


Enet4:
I have readjusted this PR to the branch head. @fredericojssilva please test and merge.

bastiao:
This branch is now merged with dev. (Also with webapp)

PR #62, opened 2015-06-11
Update search servlet to support additional fields

### LINKED ISSUE #15: Search-Webservice: Additional options
These options could be added to the Search service in order to make it more fexible.

* fields:
    * Select which fields will be returned in the query.
    * The default could be the DICOM DIM fields.

Enet4:
Additional fields are introduced with the "field" query string. (e.g. `field=RequestingPhysician&field=SOPInstanceUID`)
Fixes #15.


bastiao:
@tmgodinho can you give a quick overview of the implementation?

tmgodinho:
This should fix the problem. However, is it really necessary to do that parsing stage just to allow values to be separated by ';'?
I makes me wonder whether the other servlets also conduct this parsing stage, and if it is really worth the added complexity just for the sake of allowing values to be separated by ';'.
Moreover, I wonder if jetty is not doing this kind of parsing by himself.

bastiao:
Well, that may be a relevant question if we have for instance a large number of fields. The query string will be larger and consequently could be rejected in some browser due to the limit of query string. 
The alternative for this will be convert this method to POST. But maybe is something to think about in next version.

Enet4:
Yes, I purposely exposed the comma-separated interface variant in order to reduce the size of the query string.

Enet4:
Ok, I cleared the semi-colon separation to become conformant. In the future, we may add a POST operation that accepts a larger list of fields.

bastiao:
Thanks @Enet4.

Good job! I'm going to merge.

PR #76, opened 2015-06-14
Provider list #64

### LINKED ISSUE #64: webapp: list of providers are static and need to be dynamic

fredericojssilva:
Fix #64 issue


bastiao:
There are pretty nice console.logs :) it should be removed before merge. :)

Enet4:
I think the list of providers should be a multi-select, rather than either a specific one or all of them. Besides, you do not need a "provider=all" in the QS, the default behavior would be to query all of them.

As for the logs, there really are too many of those. I just chose to clear them all from the production code in #75 .

bastiao:
Post-poned to #96

Issue #83, opened 2015-06-17
Logos in login screen over the form [small screens]


Enet4:
In small screens (mobile devices), the logos at the bottom may get in the way and prevent the user from filling the login form.

![screenshot_2015-06-17_16-53-25](https://cloud.githubusercontent.com/assets/4738426/8212050/832c55b6-1511-11e5-935d-0a3f64691669.png)

PR #85, opened 2015-06-22
fix login view for mobile devices

fredericojssilva:
Fixed login logos over the form #83


Enet4:
This new login screen introduced some annoying scroll bars in the desktop version. I say we correct this after we take measures on higher priority issues.

Enet4:
This branch has gone too far behind to be useful. I propose we close this and bump #83.

PR #86, opened 2015-06-24
Make Advanced search button toggle caption on click

### LINKED ISSUE #77: [webapp] Advanced Search button is not responsive
The "Advanced" button in the search view behaves as expected, but it does not make clear to the user that clicking again reverts the search to a basic search. So I suggest that clicking this button should toggle between the captions "Advanced" and "Basic".

Enet4:
- added responsive advanced search button with toggle-like caption change ("Advanced" when in basic search mode, "Basic" when in advanced search mode).
Fixes #77.


bastiao:
To be honest, I believe that it is only a matter of style. It is a bit weird to see interface changing, but it will be changed when we change to webcore, right?

PR #89, opened 2015-06-30
Redefine default fields in search servlet

### LINKED ISSUE #88: [Search] searchDIM service lacking some properties in results
The search servlet, when in Patient mode, is not showing all information right now. Many fields such as PatientSex and Institution appear to come out blank from the web service. This is strange, because it used to work correctly not long ago.

screenshot_2015-06-25_11-03-21

Enet4:
Fixes #88. Some fields listed on the search interface were not being requested at the servlet side. I also adjusted the in-memory generic DIM patient table structure to index patients by ID rather than name.


bastiao:
Unfortunately the aggregation by patient ID does not work pretty well in real environment. There are many devices that does not use them. However a combination of both, can be a good choice. If patient ID exists, we use it, otherwise, we use patientname.

Issue #97, opened 2015-07-28
Configuration: define the domains to allow in the configuration file 


bastiao:
Related to #78

PR #99, opened 2015-08-03
CORS support with control of allowed origins

### LINKED ISSUE #74: Handle CORS properly
Many of our servlets are currently accepting all origins, which is not safe. This should be carefully reconsidered, and we may start by removing all "Access-Control-Allow-Origin" header assignments in the servlets.

Enet4:
Fixes #74 and #97.

- Created a servlet filter `CORSFilter` and applied it to the existing web services (all but web app)
- Added `allowedOrigins` property in web server settings (previously `accessControlAllowOrigins`).
- Configured loading and saving `allowedOrigins` property.

Configuration is done in the config.xml file:

```xml
<server enable="true" port="8080" allowedOrigins="..." />
```


bastiao:
Good work! :+1: 
(only few minor fixes)

PR #129, opened 2015-09-25
Added capability to have DICOM public nodes and with a description.

bastiao:
- There are DICOM nodes that are only used for internal use and should not appear in the interfaces.
- There are DICOM nodes that does not have a friendly description and so, a description was added.

For now, there is no need to bring that changes to the main interface, but this was a feature are required to external components better management the DICOM nodes.

Why master? 
- Required and urgent improvement to have.


Enet4:
Look ok. I just made `MoveDestination` immutable again, to prevent unexpected inconsistencies. For me it's ready to merge now.

PR #153, opened 2015-11-04
Add shutdown system hook so that Dicoogle terminates plugins at shutdown

### LINKED ISSUE #149: Shutdown plugins when terminating the server
Although we have a Dicoogle plugin set interface exposing a [shutdown()](https://github.com/bioinformatics-ua/dicoogle/blob/dev/sdk/src/main/java/pt/ua/dicoogle/sdk/PluginSet.java#L115) event method and a [shutdown()](https://github.com/bioinformatics-ua/dicoogle/blob/dev/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java#L260) method in our plugin controller, none of these are actually being called. Maybe it was, back in the remote GUI version, but not anymore. We should probably hook this to a termination signal.

Enet4:
Fixes #149.


bastiao:
I tried this solution https://github.com/DjDCH/Log4j-StaticShutdown but unfortunately it does not works nice with Sentry.

Enet4:
I've resolved the conflicts in this one. Maybe we should try fixing the bug again?

bastiao:
yes, agree

PR #172, opened 2015-12-04
Fixed #165: at and resolveURI with optional parameter

### LINKED ISSUE #165: StorageInterface: "at" method should also contains optional arguments
In the same way, of storage, we may want for instance to apply a decompressor. So the SDK should have this optional parameters to handle that requests that could be applied by that layer.

It was already discussed the same for store and index methods.

bastiao:


dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
```diff
@@ -329,7 +329,7 @@ public void startPlugin(String pluginName) {
      * @param location
      * @return 
      */
-    public Iterable<StorageInputStream> resolveURI(URI location)
+    public Iterable<StorageInputStream> resolveURI(URI location, Object ...args)
     {
         Collection<StorageInterface> storages = getStoragePlugins(true);
```
Enet4: The args in the call to the resolved store in `resolveURI` is missing.

PR #174, opened 2016-01-10
Added Patient ID in slot and added StudyInstanceUID to be returned in searchDIM

bastiao:
1. Added PatientID in the slot. It is just an extra field that may be useful.
2. Added StudyInstanceUID otherwise it will not be possible to fetch  the uid in study level

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
-node_modules
+node_modules/
 dist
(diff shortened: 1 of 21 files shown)


Enet4:
Thanks. I have cherry-picked your commits into the branch.

PR #177, opened 2016-01-14
Bug/fix logout

bastiao:
It fixes the logout. Apparently there are issues with transictionTo and this commit change to a different implementation:

http://stackoverflow.com/questions/32033247/react-router-transitionto-is-not-a-function


Enet4:
I tweaked the App component and added the cursor change to a pointer when hovering the log-out button. It's good. :+1:

PR #178, opened 2016-01-14
Fix about dimensions

bastiao:
- Fix about dimensions
- Fix license issue of react


Enet4:
Looks a bit "stretchy" on my screen (the new width makes the text harder to read). Perhaps simply using a smaller width and centering the view would be preferred.

And since we're looking for accuracy now, let us move "flux" away and put "reflux", which is what we're actually using.

bastiao:
Good. Thanks.

PR #179, opened 2016-01-14
- Expose more packages by browsify to be accessed by WebPlugins and a…

bastiao:
- Expose more packages by browsify to be accessed by WebPlugins and avoid duplicate dependencies. 
- Make the Plugins (result-entry) to appear inline


Enet4:
Looks ok. Maybe we should consider integrating a task runner, as our scripts are starting to look uglier and harder to maintain.

bastiao:
yes, you are right. And lot of replicate "code". I will merge this one, and we can open an issue for that.

PR #187, opened 2016-01-26
RBAC + Restrictions to admin/non admin and restrictions for each web plugin 

bastiao:
Dependency:

https://github.com/bioinformatics-ua/dicoogle-client-js/pull/3/files


dicoogle/src/main/java/pt/ua/dicoogle/plugins/webui/WebUIPlugin.java
```diff
@@ -75,6 +82,16 @@ public static WebUIPlugin fromPackageJSON(JSONObject obj) throws PluginFormatExc
             plugin.slotId = objDicoogle.getString("slot-id");
             plugin.moduleFile = objDicoogle.optString("module-file", "module.js");
             plugin.caption = objDicoogle.optString("caption", null);
+            JSONArray rolesArr = objDicoogle.getJSONArray("roles");
```
Enet4: The "roles" field should be optional. Please do not assume it's always there.

PR #191, opened 2016-01-29
Improve Dicoogle webapp and Dicoogle client usage

Enet4:
`dicoogle-client` 3 is up. This PR forces the use of this version of Dicoogle client (rather than the pre-release) into the webapp and performs a few more tweaks and fixes that were necessary.

- Use dicoogle-client@^3.0.2
- Fixed Restlet services returning 405 when receiving an OPTIONS request
- Prevented bad search results in search servlet from throwing NPE
- Updated linting rules and documentation on building the webapp (includes solution to error related to dicoogle-webcore)
- Cleared linting warnings
- Added TODO entries where the Dicoogle client can be used directly in the future


bastiao:
You forget to upgrade webcore to 3.0.2.
Just did and going to merge

Enet4:
All right, feel free to update that. It wouldn't matter that much anyway, ^3.0.0-rc0 also resolves to 3.0.2 and only our webapp is using webcore.

bastiao:
I did it because you also did in webapp, so it make sense to keep them up to date and sync at this level.

PR #215, opened 2016-03-30
SOPList updated - support for VLWholeSlideMicroscopyImageStorage

rlebre:


Enet4:
After a more thorough look into the code, I'm afraid that including the UID there is not enough. The [`SOPClassSettings`](https://github.com/bioinformatics-ua/dicoogle/blob/master/dicoogle/src/main/java/pt/ua/dicoogle/server/web/management/Dicoogle.java#L468) constructor should also be updated to include the name of that SOP class. You may wish to update this branch.

bastiao:
@Enet4 you are right. We need also to add UID.BreastTomosynthesisImageStorage, which we are also supporting.

PR #232, opened 2016-05-31
Search service: optimize and support pagination

Enet4:
- Optimized memory usage of `DIMGeneric` and `SearchServlet` by avoiding in-memory JSON representations.
- Added new parameters `psize` (page size) and `offset` to `/search` in order to have pagination.
   - When provided, the length of `results` will be at most `psize`, and the first `offset` results are not sent.
   - `numResults` still shows the total number of results obtained from the search (it is not clamped when paging)
- Added parameter `depth` to `/searchDIM` in order to restrict rendering certain levels of the DIM hierarchy.
   - `depth` can be either "none", "patient", "study", "series" or "images" (default)
   - "none", although apparently useless, can be used to retrieve only the number of patients.
- Adjusted search service output to always be valid JSON, even in case of error.

screenshot_2016-06-01_00-30-22

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/core/dim/DIMGeneric.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/core/dim/DIMGeneric.java
@@ -19,10 +19,14 @@
 package pt.ua.dicoogle.core.dim;
 
 
+import java.io.IOException;
 import java.io.StringWriter;
+import java.io.Writer;
 import java.net.URI;
 import java.util.*;
 
+import org.json.JSONException;
+import org.json.JSONWriter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


Enet4:
I've just added a few more improvements:
- Refactored `DIMGeneric` to avoid unnecessary aggregations on construction (when `depth` is sufficiently low)
- Patient-level pagination now works with the same `psize` and `offset` parameters (e.g. `psize=2` will show the information of two patients). Like before, it does not affect `numResults`
- added `elapsedTime` to the output of `/searchDIM`

It appears to work fine, but if we find it to be unstable, we can rewind to the previous commit.

bastiao:
- it should be tested with large datasets, just to be sure that does not broke.
- also DICOM services uses the same class, and C-FIND + C-MOVE + C-STORE needs to be tested very carefully because they are critical services.

PR #235, opened 2016-06-06
Fix webapp export source

Enet4:
- Updated webapp export operation to only use the providers selected in the last search.
- Also tweaked the implementations of related servlets.


bastiao:
Until now: problem with providers. Webapp send "providers[]"

Enet4:
I fixed that with 4d8cb37. This should be ready to merge.

PR #236, opened 2016-06-07
Support for modifying DICOM service ports in the web app

Enet4:
This is the missing piece of #212.

- Refactored DICOM service management form rows into a separate component (`ServiceForm`)
- Added validation to the service port input
- Effective support for changing the port (by pressing Enter on the port field)
- Also added a loading animation while a settings change is issued


bastiao:
I'm not able to test the branch. See: 
problem-services

Enet4:
@bastiao Try updating your npm packages before building the webapp again. You might be using a version of `react-bootstrap` older than `0.29.4`.

Enet4:
Nevertheless, that made me find another bug in the code. It's fixed in 6635fba.

bastiao:
You added in the PR:
- logs
- netbeans configurations 
- file named webapp (probably symlink)
- other staff server-settings.xsd ? 

We need to clean up these files before merge

Enet4:
That was definitely unintended.... I removed them now.

PR #237, opened 2016-06-09
UI improvements (icons, and size)

bastiao:
- Increase pagination size
- Added Icons to Side Bar 
- Reduce the size of tables (line height)


Enet4:
It seems that the plugin menu items are horizontally misaligned with the remaining items. Will you fix this?
screenshot_2016-06-09_11-07-57

Other than that, it looks good. :+1:

PR #238, opened 2016-06-09
Search and Results in the same page. 

bastiao:


Enet4:
It's looking good! Just a small nitpick: pressing Enter to search doesn't hide the list of suggested attributes.

It's also fine that you deprecated `SearchResultView`. This also reminds me: are we ever supporting routes to a specific search? (e.g. navigating to `demo.dicoogle.com/#/search?query=CT`)

bastiao:
There was three routes that should be supported:
- search with query (not working properly)
- dump of specific sop instance uid (works)
- image of a specific sop instance uid (works)

PR #240, opened 2016-06-11
Improvement UI of wizard steps 

bastiao:
Note: there are still missing number of studies, series and images (that I will be added soon). Meanwhile, it is already open to comments.

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/server/web/webapp/WEB-INF/js/stores/searchStore.js
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/server/web/webapp/WEB-INF/js/stores/searchStore.js
@@ -30,7 +30,7 @@ const SearchStore = Reflux.createStore({
           self._contents = data;
 
           //DEBUG WAIT
-          setTimeout(self.triggerWithDelay, 700)
+          setTimeout(self.triggerWithDelay, 50)
         },
         function(xhr){
           //FAILURE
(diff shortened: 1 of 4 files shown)


bastiao:
Also, noticed a bug a bug related with search: 
if you are in the study, series or image step, and try to search again, it takes a while in loop, probably some component state that need to be updated before the trigger onSearch.

dicoogle/src/main/java/pt/ua/dicoogle/server/web/webapp/WEB-INF/js/stores/searchStore.js
```diff
@@ -30,7 +30,7 @@ const SearchStore = Reflux.createStore({
           self._contents = data;
 
           //DEBUG WAIT
-          setTimeout(self.triggerWithDelay, 700)
+          setTimeout(self.triggerWithDelay, 50)
```
Enet4: Maybe we could clear the timeout entirely? The need for it might have emerged from other bad practices that we've most likely cleaned up by now.

bastiao: yes, after other priorities (fixes) it will be cleaned.

Enet4:
This might be asking too much for a small nitpick, but can the DIM level buttons' click window be expanded a bit more? For example, clicking where the mouse cursor is in the picture below won't trigger a level change, it has to be closer to the label.
screenshot_2016-06-12_21-05-56

bastiao:
all the reported issues are now fixed

PR #241, opened 2016-06-13
Human-readable elapsed time in task status

Enet4:
Before:
screenshot_2016-06-13_13-56-45

After:
screenshot_2016-06-13_14-20-19


bastiao:
Good job! :+1:

PR #244, opened 2016-06-27
Bug/export to csv

tmgodinho:
Fixes escaping in the csv servlet

Fixes bug when downloading exported csv files in Mozilla Firefox browser


dicoogle/src/main/java/pt/ua/dicoogle/server/web/webapp/WEB-INF/js/stores/exportStore.js
```diff
@@ -60,9 +60,14 @@ const ExportStore = Reflux.createStore({
           const response = JSON.parse(data);
           console.log("UID:", response.uid);
           const link = document.createElement("a");
+          const hacked_footer = document.getElementById("hacked-modal-footer");
+          link.style.visibility = "hidden";
           link.download = "file";
           link.href = Endpoints.base + "/exportFile?UID=" + response.uid;
+          hacked_footer.appendChild(link);
+          console.log(link);
```
Enet4: Please avoid adding `console.log` lines unless truly necessary.

dicoogle/src/main/java/pt/ua/dicoogle/core/query/ExportToCSVQueryTask.java
```diff
@@ -103,11 +104,16 @@ private void printLine(SearchResult result){
...
 			Object temp1 = extraFields.get(tag);
-
-			String s = (temp1 != null) ? StringUtils.trimToEmpty(temp1.toString()) : "";
+			String temp1String = temp1.toString();
+			
+			if(NumberUtils.isNumber(temp1String)){
+				temp1String = NumberUtils.createBigDecimal(temp1String).toPlainString();	
+			}
+			
+			String s = (temp1 != null) ? StringUtils.trimToEmpty(temp1String) : "";
```
Enet4: If `temp` could be null at this point, it would've already raised an exception in line 110. You probably mean to check this exception before attempting to test for a number.

dicoogle/src/main/java/pt/ua/dicoogle/core/query/ExportToCSVQueryTask.java
```diff
@@ -103,11 +104,16 @@ private void printLine(SearchResult result){
...
-
+		
 		for (String tag : tagsOrder) {
 			Object temp1 = extraFields.get(tag);
-
-			String s = (temp1 != null) ? StringUtils.trimToEmpty(temp1.toString()) : "";
+			String temp1String = temp1.toString();
+			
+			if(NumberUtils.isNumber(temp1String)){
+				temp1String = NumberUtils.createBigDecimal(temp1String).toPlainString();
```
Enet4: Do we really need to print numbers without an exponent field? What if they are extremely large/small in absolute value?

PR #256, opened 2016-10-04
Clean up old gfx and anonymouspatient artifact

Enet4:
- remove pt.ieeta.anonymousPatient artifact, for it's not being used in current code (some dead code is removed in the process) and is conflicting with new plugins.
- remove the gfx folder containing graphics from the RMI client application.
- set user role tests to be ignored


bastiao:
Good :+1: 
Merged with dev.

PR #258, opened 2016-10-04
Complete search result counters

Enet4:
- add missing counters for number of studies, series and images
- tweak search result Step container to become stateless
- do minor tweaks to sass definitions related to step bar

screenshot_2016-10-04_18-44-45


bastiao:
Merged with dev branch 👍 .

PR #262, opened 2016-10-21
Plugin controller guards

Enet4:
Added try-catch guards at the borderline of most plugin methods (i.e. unindex and many others).


bastiao:
We are using some test cases for validation (before merge). When it is ready, we will merge.

PR #276, opened 2017-01-05
Webapp - AETitle field size 

bastiao:
1) Allow different size on AETitle size.
2) Just make the standard button


Enet4:
Look ok. Would you like to just include a [warning](https://react-bootstrap.github.io/components.html#forms-validation) for non-compliant AE titles? Also, are some systems case sensitive? If so, then we should not lock capitalization either.

Enet4:
I've detected a few problems while looking into the aforementioned warning. I'll just push them directly to this branch.

- "btn_dicoogle" should not go to `bsStyle`, but to `bsClass` (we're overriding the component's classes). The "btn" class was also missing for the button to look ok.
- The cancel button was also adding the server. This might have been something we overlooked from a previous contribution.

bastiao:
@Enet4 sure! I will give you feedback soon

Issue #63, opened 2015-06-11
Improve the deployment approach (mvn build/install) 


bastiao:
By default, including all node_modules dependencies is quite heavy and time spending.

Enet4:
`devDependencies` are only needed for building the application, whereas the other ones are required for the deployment. I've seen some dependencies in the web app being incorrectly assigned as common dependencies when they are actually dev dependencies.
Are you suggesting to include the web app's building script into the maven processes?

bastiao:
yes, it is one possibility, with grunt plugin or whatever. But for now, we need to evaluate the best way to do it.

Enet4:
We may discuss this tomorrow. But since we have a prepublish script,
performing an npm install from maven is enough to build it. Deploying it
would then involve pruning the dependencies.
And it seems a maven plugin is available.
https://github.com/eirslett/frontend-maven-plugin

Eduardo Pinho
http://enet4.wordpress.com

2015-06-12 0:07 GMT+01:00 Luís Bastião <notifications@github.com>:

Enet4:
The [frontend-maven-plugin](https://github.com/eirslett/frontend-maven-plugin) requires that Node.js and npm are installed locally. I find it overkill for development purposes, but then it would work on any machine with Maven.
Alternatively, we could just use mojo to execute `npm`. So what do you think? And by the way, how should we separate the Java source building process with the web app's?

PR #278, opened 2017-01-06
Improve the webapp deployment approach 

bastiao:
As discussed in #63 added a better way to build the webapp. It allows to install node and npm, and also to build the webapp for production. 

For build you just need to run `mvn install`. Alternative, and you may want, skip the npm just do:

`mvn install -Dskip.npm`


Enet4:
Well, as I recall from #63, I still uneasy for me that I would have to install a "pocket" version of Node.js and npm in order to use that script, even when it's already installed in my system. Ideally, we would have a way to detect whether supported versions of `node` and `npm` are available (maybe by calling `npm --version`) and use those instead.

On the other hand, I can simply choose to skip that step and build the webapp manually. I'm a bit torn, because even something as simple as a shell script would suit me.

One more remark: is the documentation properly documented? Step 3 of building Dicoogle still states that you must navigate to the webapp and run `npm install`.

bastiao:
@Enet4 fully agree. We need to sustain the deployment. If you want to keep developing, skip this installation is just an easier way to go.
Just fixed the README.md, thank you.

README.md
```diff
@@ -135,11 +135,10 @@ If you are in the development process and using the UI, you may need to change y
 
 
  1. Retrieve the full source code from this repository: `git clone https://github.com/bioinformatics-ua/dicoogle.git`
- 2. Navigate to the web app directory: "./dicoogle/src/main/resources/webapp/"
- 3. Build the web application: `npm install`
-    - The full instructions can be read [here](dicoogle/src/main/resources/webapp/README.md).
- 4. Navigate to the project's base directory, and build the parent Maven project by calling `mvn install`.
- 5. The resulting jar file can be found in "./dicoogle/target".
+ 2. Navigate to the project's base directory, and build the parent Maven project by calling `mvn install`.
+    - Note: if you want, you can skip the npm part: `mvn install -Dskip.npm`
```
Enet4: I think it's not clear that this part refers to building the webapp. Maybe we should either extend this note or add a second list of steps to do it manually.

bastiao: sure, that's well be clarified in the pending PR regarding README. Thank you

tmgodinho:
I think that this is a cool approach if it works well under windows.

bastiao:
it works in windows, indeed.

Issue #257, opened 2016-10-04
Deprecate StorageInterface#handles


Enet4:
By definition, a certain storage is capable of handling a file with the given URI if and only if the scheme associated to the storage is equal to the file URI's scheme. Giving plugin developers the "opportunity" to change this behaviour via the [`handles`](https://github.com/bioinformatics-ua/dicoogle/blob/master/sdk/src/main/java/pt/ua/dicoogle/sdk/StorageInterface.java#L48) method is not only useless, but also an open door for bugs in plugins.

I propose that we replace this method declaration with the following code in the next major version, advising plugin developers not to use or override it.

``` java
@Deprecated
default public boolean handles(URI location) {
    Objects.requireNonNull(location);
    return Objects.equals(this.getScheme(), location.getScheme());
}
```

jorgeMFS:
I agree. Due to this issue, I was having a StackOverFlowError due to my erroneous implementation.

`
public boolean handles(URI location) {
		return Objects.equals(getScheme(), this.AnonymousScheme);
	}
`

PR #280, opened 2017-01-26
Update SDK with default methods and Java 8

Enet4:
These changes will allow us to expand the plugin set without future breaking changes, as well as make future plugin implementations simpler. For instance, developers will no longer need to implement every single getter method, including those that we no longer support. 

Possible breaking changes:
 - Java 8 is required. This is fine, since Java 8 has been stable for quite a while, and we no longer feel the need to support Dicoogle running over Java 7.
 - If a plugin was using reflection to fetch `PluginSet` instances, attempting to get the plugins within might lead to a runtime exception. Since AFAIK no one has tried doing this, it should not become a problem.

Additional changes:
 - Method prototypes in PluginSet getters are slightly modified to return collections of `? extends Plugin` (this fixes some inconsistencies that led to casts in our implementations)
 - Many methods in PluginSet and some plugin interfaces are given a default implementation: getters in PluginSet return empty collections of plugins, `shutdown()` does nothing, `handles()` in indexers return true, and `handles()` in storage interface was marked as deprecated (according to #257).
- Getters for graphical plugins in `PluginController` were also simplified to return empty lists.

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
@@ -240,7 +240,7 @@ public class PluginController{
 
         ArrayList<ServerResource> restInterfaces = new ArrayList<>();
         for (PluginSet set : plugins) {
-            Collection<ServerResource> restInterface = set.getRestPlugins();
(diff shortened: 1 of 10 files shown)


bastiao:
Let's jump to the unstable era and make QA working hard.

Issue #270, opened 2016-11-24
dic2png cached image mapping fails for some URIs


Enet4:
Certain URIs containing special characters not allowed by the supporting file system will make our DICOM-to-PNG service fail, because it cannot save the resulting images. We should improve how URIs are mapped in the image cached, in a way to make it resilient.

PR #283, opened 2017-01-30
Fixed bug when trying to visualize thumbnails on Windows.

bastiao:
If you are use Windows systems, using the classic approach in URI it may be returned  ":". This is a hot fix to avoid this kind of problems.


Enet4:
Looks ok. We have created #270 very likely because of this particular problem. :)
On the other hand, I suggest we keep #270 open, so as to consider making a more robust solution in the future.

Issue #266, opened 2016-11-07
Server settings redesign


Enet4:
Many of our future issues (e.g. #98 #122 and #255) will require more core server configurations. Given the current, difficult state of our server settings design and implementation, this issue is a proposal for a redesign and reimplementation of dicoogle's server settings (what is currently known as config.xml).

The new format is still based on XML, but with a different, more manageable schema.

You might find that some fields are missing. They were either deemed redundant or just made of obsolete settings that no longer belong in the core settings.

I could be missing some important fields though, so feedback is important.

We are also refactoring the programmatic API (see the new [`ServerSettings`](https://github.com/bioinformatics-ua/dicoogle/blob/imp/server-settings-refactor/sdk/src/main/java/pt/ua/dicoogle/sdk/settings/server/ServerSettings.java) and [`ServerSettingsReader`](https://github.com/bioinformatics-ua/dicoogle/blob/imp/server-settings-refactor/sdk/src/main/java/pt/ua/dicoogle/sdk/settings/server/ServerSettingsReader.java)). We'll be using a setter and a getter for each property, with methods for reaching each configuration namespace. Currently, there are 3 namespaces, one of them with 2 namespaces: "web-server", "archive", and "dicom-services" (containing "storage" and "query-retrieve").

You may find a configuration example [here](https://github.com/bioinformatics-ua/dicoogle/blob/imp/server-settings-refactor/dicoogle/src/test/resources/pt/ua/dicoogle/core/settings/test-config-new.xml). It's currently being used for testing too. :)

Proof of concept: https://github.com/bioinformatics-ua/dicoogle/tree/imp/server-settings-refactor

tmgodinho:
Is it really necessary to change the file format for the configurations??
Some of our users are not familiar with JSON. Moreover, they have spent considerable time aquiring expertise on XML.
I propose keeping the configurations file in XML.

Enet4:
> Is it really necessary to change the file format for the configurations??

I have shown the benefits above. We use the same file's base name (as in, only change the extension) and fall back to the old settings (if no "config.json" is found) until the following major version, in which we'd remove the old settings model. Also note that it's not much a big deal that we're changing the format, considering that we'll have to change the implementation anyway.

> Some of our users are not familiar with JSON. Moreover, they have spent considerable time aquiring expertise on XML.

[...]

PR #284, opened 2017-01-31
Server settings refactor

Enet4:
Discussion from #266 continues here. This is the latest implementation of the new server settings. Since it's a breaking change, it may remain on hold until we are sure that it's safe to merge.

Summary of changes:

- [`ServerSettings`](https://github.com/bioinformatics-ua/dicoogle/blob/imp/server-settings-refactor/sdk/src/main/java/pt/ua/dicoogle/sdk/settings/server/ServerSettings.java) is no longer a singleton, but actually an interface for reading and writing all server settings (it extends `ServerSettingsReader`, which is the read-only interface).
- The two interfaces mentioned above have a new API, structured as a hierarchy (breaking change).
- Access to a managed server settings instance is now `ServerSettingsManager.getSettings()`, saving settings is `ServerSettingsManager.saveSettings()`.
- Added a new implementation for server settings based on Jackson, which follows a new XML schema.
- New settings are saved in a new folder dedicated to configuration files "confs". 
- Added migration from old format to the new one on startup.
- Another great portion of rGUI code was cleansed.

Right now we should continue testing this. I also might still have overlooked some useful properties.


bastiao:
@Enet4 are we going forward with this approach? It is tagged as "in progress" but should be it be stable to at least go to QA?

bastiao:
They have two different flavours, progress or/and discussion. But I got your point. 
Anyway, considering the abstention, we need to move forward anyway, seems to a good time to look, test and merge the PR.

Enet4:
Ok, let us look into this in the near future.

PR #285, opened 2017-02-02
Remove not used dependency (email report).

bastiao:
Just remove some dead code. We do not need this dependency anymore and there is no need to carry it.


Enet4:
This PR looks ok, but it seems that there's a problem with the base branch. The class `pt.ua.dicoogle.sdk.settings.Utils` attempts to use `pt.ua.dicoogle.sdk.settings.types.GenericType`, which currently does not exist at all! I wonder what happened...

bastiao:
Thanks for comment, but it is offtopic I guess.

PR #291, opened 2017-03-27
Allow to change the number of tasks by environment propriety.

bastiao:
You can change the number of parallel tasks by running:

```
java -Ddicoogle.taskManager.nThreads=8 -jar dicoogle.jar
```

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/server/ControlServices.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/server/ControlServices.java
@@ -50,7 +50,7 @@ public class ControlServices implements IServices
     
     private ControlServices()
     {
-        TaskManager taskManager = new TaskManager(4);
+        TaskManager taskManager = new TaskManager(Integer.parseInt(System.getProperty("dicoogle.taskManager.nThreads", "4")));
 
         startInicialServices();
     }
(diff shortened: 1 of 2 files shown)


dicoogle/src/main/java/pt/ua/dicoogle/server/ControlServices.java
```diff
@@ -50,7 +50,7 @@
     
     private ControlServices()
     {
-        TaskManager taskManager = new TaskManager(4);
+        TaskManager taskManager = new TaskManager(Integer.parseInt(System.getProperty("dicoogle.taskManager.nThreads", "4")));
```
Enet4: This looks superfluous. Don't we want a single task manager? This one is just dropped out of scope when leaving the constructor.

bastiao: True. I just did not want to do major refactor at this time.

Enet4: Removing this line should be nothing more than an improvement, not a refactor. :)
If it breaks something, for some reason, we can keep it and address the problem separately.

bastiao: Yes, that's a good idea to address the subject separately 👍

PR #293, opened 2017-04-13
Remove dangling task manager in ControlServices

Enet4:
As detected in #291.


bastiao:
Right code, wrong place.

PR #295, opened 2017-09-19
Fixed "java.io.FileNotFoundException: roles.xml 

### LINKED ISSUE #265: Create roles.xml on startup if missing
A new Dicoogle deployment will print a stack trace because of a missing roles.xml, whereas other configuration files are automatically created. We should make Dicoogle create a barebones file instead.

Example roles.xml:

```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<Roles>
  <Role name="pen" />
  <Role name="healthcare" />
</Roles>
```

rlebre:
Fixed exception on startup when there is not any file called "roles.xml" (fixes #265 )

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/server/users/RolesXML.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/server/users/RolesXML.java
@@ -101,11 +101,17 @@ public class RolesXML extends DefaultHandler
     public RolesStruct getXML()
     {
         roles.reset();
-        
-        try
-        {
 
-            FileInputStream fin = new FileInputStream("roles.xml");
+        FileInputStream fin;
+
+        try {
+            fin = new FileInputStream("roles.xml");
+        } catch(FileNotFoundException e){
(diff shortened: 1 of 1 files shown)


jorgeMFS:
Uau

Em 19/09/2017 18:59, "Rui Lebre" <notifications@github.com> escreveu:

dicoogle/src/main/java/pt/ua/dicoogle/server/users/RolesXML.java
```diff
@@ -101,11 +101,17 @@ private String resolveAttrib( String attr, Attributes attribs, String defaultVal
...
     {
         roles.reset();
-        
-        try
-        {
 
-            FileInputStream fin = new FileInputStream("roles.xml");
+        FileInputStream fin;
+
+        try {
```
Enet4: I believe one could also have a single try statement and catch `FileNotFoundException` at the end, next to the other catch blocks. But this seems fine.

PR #296, opened 2017-09-29
Change list of providers to multiselect

### LINKED ISSUE #96: Change list of providers to multiselect
Related with #76 

> the list of providers should be a multi-select, rather than either a specific one or all of them

funcionante:
Resolves #96.

### DIFF
--- a/dicoogle/src/main/resources/webapp/index-template.html
+++ b/dicoogle/src/main/resources/webapp/index-template.html
@@ -30,14 +30,17 @@
 
     <link rel="stylesheet" href="css/jquery-ui.css">
 
-     <!--Custom JS -->
-     <!--
+    <!-- React Select CSS -->
+    <link rel="stylesheet" href="https://unpkg.com/react-select/dist/react-select.css">
+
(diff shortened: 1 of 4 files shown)


dicoogle/src/main/resources/webapp/index-template.html
```diff
@@ -30,14 +30,17 @@
 
     <link rel="stylesheet" href="css/jquery-ui.css">
 
-     <!--Custom JS -->
-     <!--
+    <!-- React Select CSS -->
+    <link rel="stylesheet" href="https://unpkg.com/react-select/dist/react-select.css">
```
Enet4: We prefer keeping these resources locally. Can you add this css file to our css/ folder and import it from there?

bastiao:
Nice change. We hope that react-select release a final/stable version soon (instead of RC - release candidate).

bastiao:
react-select has now final releases. It might be interested to upgrade.
https://github.com/JedWatson/react-select/releases

funcionante:
@bastiao, now it is upgraded.

Issue #249, opened 2016-07-26
Webapp using old authentication token before logging in to a fresh server


Enet4:
Not long ago, we merged #242 so that people could easily choose to login with a different account when navigating to the root path. That part seems to be fine, but there appears to be some inconsistency with the authentication tokens when logging in for the first time in a fresh server.

My browser had a token in cache from a different server session (the server was restarted since this token was generated). For some reason, after re-entering the credentials of the same user, the web plugin menus were missing. I checked the requests made to `/webui` and they were performed even before `/login`, so it used the old token. I had to refresh the page to make the application use the new one when retrieving the web plugins.

I can only reproduce this by restarting the server and entering it at the root of the webapp while another token is already cached in the browser.

Enet4:
I believe this is already fixed in #298.

PR #298, opened 2017-10-01
Refactor login and user session handling to address login issues

Enet4:
This PR fixes issues in the user store and related pages that was resulting in multiple HTTP login requests and errors when attempting to load a session in local storage.

@rlebre: Can you check whether this fixes or helps on menu webplugins not being loaded properly (#249)? I haven't tested this for users in multiple roles yet.


rlebre:
As far as I tested, it seems like this pull request fixes #249 (not tested in multiple roles) with menu webplugins being loaded right as expected

bastiao:
The first impact with the page:  Login Failed. Please try again. 

Any idea why?

(Tested with Firefox)


firsttime

Enet4:
I'm afraid I can't reproduce. Did the console print any errors? Is the server running? What login credentials have you tried?

A JavaScript error occurs when pressing the Login button with and empty username and password, but that's fixed in #301.

Enet4:
@bastiao Ok, I found it right now. It only happens on Firefox. I'll look into it now.

PR #299, opened 2017-10-04
Fix overlapping footer in small screen devices

funcionante:
Also fixes the unintended reveal of the content hidden by the login block. This targets the issue #83.

Before and after:
b0
b1


bastiao:
very good. 👍 

Offtopics:
1. @funcionante you can work directly in branch from the bioinformatics-ua repository 
2. @rlebre can we get the images with white background (FCT) without background (transparent)? did you did for new website?

rlebre:
@bastiao no, I didn't. I took out the QREN/FCT/UE pictures and let the text information stay. But I can get the transparent background images

rlebre:
fct
qren_ue


@funcionante @bastiao here they are

funcionante:
Thank you. I replaced with these new ones.

Issue #229, opened 2016-05-25
Improve indexation request component


Enet4:
![screenshot_2016-05-25_14-38-58](https://cloud.githubusercontent.com/assets/4738426/15541957/a1514f96-2286-11e6-839b-3565790f2048.png)

I have two useful ideas for this user interface component, that I have felt the need for recently.
- Add a multi-selection component for choosing which indexers to use.
- Provide auto-complete hints containing all registered storage schemes (e.g. `file:`, `s3:`, ...) in the directory path input bar.

Enet4:
I've increased the priority, given that many new users found this interface confusing.

PR #303, opened 2017-10-15
Add multiselect for choosing which providers to use when indexing

funcionante:
This is targeting the topic "Add a multi-selection component for choosing which indexers to use." on #229.


dicoogle/src/main/java/pt/ua/dicoogle/server/web/servlets/management/ForceIndexing.java
```diff
@@ -60,7 +60,7 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws S
 
     // Getting Parameters.
     String[] uris = req.getParameterValues("uri");
-    String[] pluginsName = req.getParameterValues("plugin");
+    String[] pluginsName = req.getParameterValues("plugin[]");
```
Enet4: We shouldn't be doing this. Modifying the server's API is a breaking change (and in this case, an unnecessary one). We should instead adjust the web client to build multiple "plugin" query string parameters (e.g. `/foo?plugin=x&plugin=y`).

dicoogle/src/main/resources/webapp/js/handlers/requestHandler.js
```diff
@@ -193,12 +193,13 @@ function saveIndexOptions(path, watcher, zip, saveThumbnail, effort, thumbnailSi
 
 }
 
-function forceIndex(uri){
+function forceIndex(uri, plugin){
   //console.log(state);
   // TODO use dicoogle client
   $.post(Endpoints.base + "/management/tasks/index",
```
Enet4: You should instead use our Dicoogle client. The [`index`](https://bioinformatics-ua.github.io/dicoogle-client-js/interfaces/_types_dicoogle_client_d_.dicoogleclient.dicoogleaccess.html#index) method already supports a list of index providers.

funcionante:
Now targets the whole issue #229.

dicoogle/src/main/resources/webapp/js/actions/pluginActions.js
```diff
@@ -0,0 +1,4 @@
+import Reflux from 'reflux';
+export const PluginActions = {
+  get: Reflux.createAction()
```
Enet4: We can export actions individually instead: `export const get = Reflux.createAction();`

dicoogle/src/main/resources/webapp/js/components/indexer/IndexStatusView.js
```diff
@@ -1,20 +1,37 @@
 import React from 'react';
 import {IndexStatusActions} from "../../actions/indexStatusAction";
 import {IndexStatusStore} from "../../stores/indexStatusStore";
+import {PluginActions} from "../../actions/pluginActions";
```
Enet4: With the changes in pluginActions.js, you would now do `import * as PluginActions from "../../actions/pluginActions";`

dicoogle/src/main/resources/webapp/js/components/indexer/IndexStatusView.js
```diff
@@ -98,6 +153,50 @@ const IndexStatusView = React.createClass({
         else{
           IndexStatusActions.stop(uid);
         }
+      },
+      _onProvidersChange: function(data) {
+        this.setState({providers: data.data});
+      },
+      _onStoragePluginsChange: function (data) {
+        var schemeList = data.data.map(plugin => plugin.scheme);
```
Enet4: Let's avoid using `var`: `let schemeList = ...`

dicoogle/src/main/resources/webapp/js/stores/pluginStore.js
```diff
@@ -0,0 +1,33 @@
...
+        self.trigger({
+          success: false,
+          status: xhr.status
+        });
+      }
+    );
+  }
+});
+
+export {PluginStore};
```
Enet4: Let's make this one a default export: `export default PluginStore;`

Then you would import it like this: `import PluginStore from ...;`

dicoogle/src/main/resources/webapp/js/components/indexer/IndexStatusView.js
```diff
@@ -28,20 +45,22 @@ const IndexStatusView = React.createClass({
          //Stop refresh interval
          clearInterval(refreshIntervalId);
 
+         this.unsubscribeProviders;
+         this.unsubscribeIndexStatus;
+         this.unsubscribePlugin;
```
Enet4: Oops! Parentheses are missing. It should be `this.unsubscribeProviders();` and the same to the other two lines.

funcionante: Oops x) Thank you!

bastiao:
Indeed, quite nice improvement in UI! 👍 
Just did the merge and quick fixes during merge - and will merge to dev after the Travis-CI builds.

Issue #263, opened 2016-10-26
Show enabled plugins on webapp


Enet4:
Extend our Plugins & Services tab to also exhibit the list of installed plugins (including query providers, indexers, stores, ...)

Enet4:
Already done in #305 (and fixed later on in #404).

PR #305, opened 2017-10-29
Manage plugins on webapp

funcionante:
This targets issue #263 and looks like the image below.
show_plugins_dicoogle
For now, plugins are expanded with its attributes when clicked, which may not be so useful, because some fields like `type` and `name` are not displayed and others are `null`. The display of this information can be better personalized when we add actions to plugins.

Depends on #310.


funcionante:
Now, with the endpoints introduced in #310, it is possible to enable or disable plugins, like this:
enable_disable_plugins_webapp

dicoogle/src/main/resources/webapp/js/stores/pluginStore.js
```diff
@@ -0,0 +1,57 @@
...
+const PluginStore = Reflux.createStore({
+  listenables: PluginActions,
+  init: function () {
+    this._contents = {};
+  },
+
+  onGet: function(type) {
+    const self = this;
+
+    request(Endpoints.base + "/plugins/" + type,
```
Enet4: Please use `dicoogle-client` directly:

```
Dicoogle.request('GET', ['plugins', type])
    .end((err, data) => { ... });
```

dicoogle/src/main/resources/webapp/js/components/management/pluginsView.js
```diff
@@ -0,0 +1,141 @@
...
+                            this,
+                            type,
+                            plugin.name,
+                            (plugin.enabled ? "disable" : "enable"))}>
+                    {plugin.enabled ? "Disable" : "Enable"}
+                  </button>
+                  {
+                    Object.keys(plugin).filter(field => ignoreFieldList.indexOf(field) < 0).map(field => (
+                      <p><b>{field}:</b> {
+                        plugin[field] !== null ? plugin[field].toString() : "undefined"
```
Enet4: Let's hide undefined fields instead, as they do not provide useful information (and some of them are just placeholders for content not yet available).

funcionante:
Done. I also changed the messages to be displayed in "toasts", like in [IndexerView](https://github.com/bioinformatics-ua/dicoogle/blob/f77cc78a1b12b6cd6a8635fe9e8e7ca68adbbfd8/dicoogle/src/main/resources/webapp/js/components/management/indexerView.js).

Enet4:
I could reproduce this, although it wasn't a trivial finding in my setup:

screenshot-2018-4-26 dicoogle

If the plugin's name is very long, there's a chance that the arrow is awkwardly placed in the next line when resizing the window. It actually looks OK if we compress the window further:

screenshot-2018-4-26 dicoogle 1

Fixing this might require a bit of CSS tweaking.

funcionante:
Well noted, I think now it is fixed.

PR #310, opened 2017-11-02
Add endpoints to enable/disable plugins in the web service

funcionante:
Endpoints:
* `POST /plugins/<type>/<name>/enable`
* `POST /plugins/<type>/<name>/disable`

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
@@ -496,7 +496,7 @@ public class PluginController{
                 return p;
             }
         }
-        logger.error("No indexer matching name {} for onlyEnabled = {}", name, onlyEnabled);
+        logger.debug("No indexer matching name {} for onlyEnabled = {}", name, onlyEnabled);
         return null;
     }
 
@@ -508,7 +508,7 @@ public class PluginController{
                 return p;
             }
         }
-        logger.error("No indexer matching name {} for onlyEnabled = {}", name, onlyEnabled);
+        logger.debug("No indexer matching name {} for onlyEnabled = {}", name, onlyEnabled);
         return null;
     }
     
--- a/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
@@ -487,6 +487,30 @@ public class PluginController{
     	logger.error("No indexer matching name {} for onlyEnabled = {}", name, onlyEnabled);
     	return null;
     }
+
+    public JettyPluginInterface getServletByName(String name, boolean onlyEnabled){
+        Collection<JettyPluginInterface> plugins = getServletPlugins(onlyEnabled);
+        for(JettyPluginInterface p : plugins){
+            if(p.getName().equalsIgnoreCase(name)){
+                //logger.info("Retrived Query Provider: "+name);
+                return p;
+            }
+        }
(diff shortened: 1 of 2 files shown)


dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
```diff
@@ -487,6 +487,30 @@ public IndexerInterface getIndexerByName(String name, boolean onlyEnabled){
...
+
+    public JettyPluginInterface getServletByName(String name, boolean onlyEnabled){
+        Collection<JettyPluginInterface> plugins = getServletPlugins(onlyEnabled);
+        for(JettyPluginInterface p : plugins){
+            if(p.getName().equalsIgnoreCase(name)){
+                //logger.info("Retrived Query Provider: "+name);
+                return p;
+            }
+        }
+        logger.error("No indexer matching name {} for onlyEnabled = {}", name, onlyEnabled);
```
Enet4: This is not severe enough to be at ERROR level. I say we keep this (and the one below) at level INFO or DEBUG.

dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
```diff
@@ -487,6 +487,30 @@ public IndexerInterface getIndexerByName(String name, boolean onlyEnabled){
...
+
+    public StorageInterface getStorageByName(String name, boolean onlyEnabled){
+        Collection<StorageInterface> plugins = getStoragePlugins(onlyEnabled);
+        for(StorageInterface p : plugins){
+            if(p.getName().equalsIgnoreCase(name)){
+                //logger.info("Retrived Query Provider: "+name);
+                return p;
+            }
+        }
+        logger.error("No indexer matching name {} for onlyEnabled = {}", name, onlyEnabled);
```
Enet4: Same concern here.

dicoogle/src/main/java/pt/ua/dicoogle/server/web/servlets/plugins/PluginsServlet.java
```diff
@@ -140,6 +140,60 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
         }
     }
 
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
+        List<String> pathParts = sanitizedSubpathParts(req);
+
+        // the format must be /<type>/<name>/[enable|disable]
```
Enet4: Doesn't `/<type>/<name>/(enable|disable)` convey the intent better? The action isn't optional. :)

dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
funcionante: I followed the behavior of the similar methods already implemented. Should I change the others too?

dicoogle/src/main/java/pt/ua/dicoogle/server/web/servlets/plugins/PluginsServlet.java
funcionante: It does :+1:

dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
Enet4: That is odd. Since this can be triggered at the user's "fault", it is unreasonable to log these occurrences at the plugin controller and call them errors. Returning `null ` at the event of the plugin not existing is part of the contract. I say we use DEBUG for this entry and replace the others on a new PR (just in case other collaborators may wish to comment on this).

PR #314, opened 2017-11-15
Update README.md with new logo + info about Learning Pack

Enet4:
[Rendered](https://github.com/bioinformatics-ua/dicoogle/blob/imp/patch-readme-learningpack/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ Contributors:
 * Eriksson Monteiro (UA.PT Bioinformatics)
 * Tiago Godinho (UA.PT Bioinformatics)
 * Jorge Miguel Silva (UA.PT Bioinformatics)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -181,6 +181,7 @@ Contributors:
 * Eriksson Monteiro (UA.PT Bioinformatics)
 * Tiago Godinho (UA.PT Bioinformatics)
 * Jorge Miguel Silva (UA.PT Bioinformatics)
+* Rui Lebre (UA.PT Bioinformatics) - [@rlebre](https://github.com/rlebre)
```
Enet4: @rlebre Can we stay consistent here? All other contributors have a GitHub profile. We should either add them all, or keep that to maintainers only.

Issue #312, opened 2017-11-06
Update dicoogle-client (version 4) and use it for all interactions with Dicoogle


Enet4:
1. Update the webapp to use the latest `dicoogle-client` package. [Version 4](https://github.com/bioinformatics-ua/dicoogle-client-js/releases/tag/v4.0.0) introduced a few changes that need to be considered.
2. Replace all AJAX requests to the Dicoogle server (often using jQuery) to use `dicoogle-client`. This package contains a well documented programmatic API with static type definitions, and automated tests with good code coverage. It should be favoured over anything else.

PR #301, opened 2017-10-10
Support for guest credentials

Enet4:
This adds build-time configured guest credentials. Use the environment variables `GUEST_USERNAME` and `GUEST_PASSWORD` to define credentials which will be automatically used in some circumstances and publicly visible. E.g. in bash:

```bash
GUEST_USERNAME=dicoogle GUEST_PASSWORD=dicoogle npm run build
```

If the user chooses to log out, these credentials are also shown below the login form:

screenshot_2017-06-14_15-33-13

PR #315, opened 2017-11-25
Update dicoogle-client to version 4 and use it

funcionante:
This targets #312. It depends on #301, that already handles login actions using `dicoogle-client` package. Also, this solves two minor bugs, that were caught along the refactoring:
* In IndexerStatusView, now tasks are being correctly stopped when this action is performed;
* In StorageView, now it allows to correctly add a storage server with no description.


Enet4:
I have rebased this branch.

Issue #221, opened 2016-05-02
Web UI Export: Provide hints for field list.


tmgodinho:
The users claimed that it was too bothersome to insert manually all the fields to be included in the exported results.

Milton suggested some sort of tick boxes, or a multi choice list to aid the user in this operation.

This was reported by Milton.

Enet4:
We could use [react-select](https://github.com/JedWatson/react-select) to make a multi-select field with a searchable list of DICOM attributes. The [live demo](http://jedwatson.github.io/react-select/) even shows a similar example.

bastiao:
@tmgodinho another suggestion that was discussed before was to have capability to save templates with predefined tags to export. I strongly believe that will make their interaction much easier.

Enet4:
Indeed, I recall that one. I find it reasonable to contemplate both facilities. For the latter however, we need to sort out some kind of account-wise storage, right?

bastiao:
yes, indeed.

tmgodinho:
Relates with this issue: #182

PR #318, opened 2017-12-03
Save and list user's presets for fields to be exported

funcionante:
For now, this targets a part of issue #221. It defines endpoints to allow using templates with predefined fields to export:
* Get the list of a user's presets: `GET /presets`
* Save a user's preset: `POST /presets?name=<name>&content=<content>`


dicoogle/src/main/java/pt/ua/dicoogle/server/users/PresetsHandler.java
```diff
@@ -0,0 +1,88 @@
...
+
+        if (!presetsDir.isDirectory()) {
+            return presets;
+        }
+
+        for (File presetFile: presetsDir.listFiles()) {
+            try {
+                // read all lines
+                String presetContent = new Scanner(presetFile).useDelimiter("\\Z").next();
+                presets.put(presetFile.getName(), presetContent);
```
Enet4: Won't this include the `.txt` extension in the keys?

dicoogle/src/main/java/pt/ua/dicoogle/server/web/servlets/search/PresetsServlet.java
```diff
@@ -0,0 +1,88 @@
...
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
+        String token = req.getHeader("Authorization");
+        User user = Authentication.getInstance().getUsername(token);
+        if (user == null) {
+            resp.sendError(401);
+        }
+
+        String presetName = req.getParameter("name");
+        String presetContent = req.getParameter("content");
```
Enet4: I forgot one more thing: could you try making an endpoint `/presets/<<username>>/<<preset-name>>` instead? The content should be sent at the request body. Users would only have access to their own presets, but administrators could have full access eventually.

funcionante:
Thanks for the feedback. As suggested, now the endpoints are:

Get the list of a user's presets: `GET /presets/<username>`
Save a user's preset: `POST /presets/<username>/<preset-name>`

Some methods already used in other servlets, like `sanitizedSubpathParts()` and `sendError()`, may be useful globally. Should I extract and put them in another place?

Enet4:
> Should I extract and put them in another place?

Yep, that's a good idea!

dicoogle/src/main/java/pt/ua/dicoogle/server/users/UserExportPresets.java
```diff
@@ -49,38 +56,34 @@ public static boolean savePreset(String username, String presetName, String pres
...
-            Writer writer = new PrintWriter(presetFile);
-            writer.write(presetContent);
-            writer.close();
-
-        } catch (Exception e) {
-            presetFile.delete();
-            return false;
+        Writer writer = new PrintWriter(presetFile);
+        for (String field: fields) {
+            writer.write(field + "\n");
```
Enet4: Minor nit:

```
writer.write(field);
writer.write('\n');
```

dicoogle/src/main/java/pt/ua/dicoogle/server/users/UserExportPresets.java
```diff
@@ -49,38 +56,34 @@ public static boolean savePreset(String username, String presetName, String pres
...
 
-        try {
-            Writer writer = new PrintWriter(presetFile);
-            writer.write(presetContent);
-            writer.close();
-
-        } catch (Exception e) {
-            presetFile.delete();
-            return false;
+        Writer writer = new PrintWriter(presetFile);
```
Enet4: Prefer try-with syntax here:

```
try (Writer writer = new PrintWriter(presetFile)) {
    // ...
}
```

dicoogle/src/main/java/pt/ua/dicoogle/server/users/UserExportPresets.java
```diff
@@ -49,38 +56,34 @@ public static boolean savePreset(String username, String presetName, String pres
             return false;
```
Enet4: I think we should allow the user to update existing presets. Let's discard this check and let the file be overwritten.

This also means we don't need the return type to be `boolean`.

dicoogle/src/main/java/pt/ua/dicoogle/server/web/servlets/search/PresetsServlet.java
```diff
@@ -65,24 +86,65 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+
+        // the format must be /<username>
+        if (pathParts.size() != 1) {
+            sendError(resp, 400, "Illegal plugin request URI: wrong resource path size");
+            return;
+        }
+
+        String username = pathParts.get(0);
+
+        // if the specified user is not the requester, he must have admin privileges
```
Enet4: s/he/they/ (_", they must have admin privileges"_). Not important, but it's best to keep users gender-neutral.

dicoogle/src/main/java/pt/ua/dicoogle/server/web/servlets/search/PresetsServlet.java
```diff
@@ -65,24 +86,65 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+        if (!username.equals(user.getUsername()) && !user.isAdmin()) {
+            sendError(resp, 401, "Unauthorized access to this user.");
```
Enet4: The error code 403 might be more accurate here: the user is authenticated at this point, but does not have permission to access the presets.

funcionante:
Now targets the whole issue #221. The changes to the webapp now allow to:
- load previously exported presets;
- save new presets (it is possible to override previous ones);
- choose from the list of existing fields, only supporting valid values;
- only export files with at least one field;
- only save presets with non-empty names;
- display a success message when a preset is saved.

It looks like this:

image

Enet4:
Whelp, we have a performance issue here. The known list of tags is pretty big, and the default select component invokes a `filter` over the current list of candidates on each text input event. In my bench over here, some of these can take 200ms. By the looks of it, we could really use an index at the tag list storage or something similar.

screenshot_2017-12-18_17-13-06

I also think that the user should be able to specify custom tags even if they are not listed, because they might be private and not yet registered in the platform by the user (AFAICT the webapp doesn't facilitate this either!). But I'd like to hear more opinions (CC @tmgodinho @bastiao )

funcionante:
Now the filtering is being done asynchronously. Also, as nobody opposed it, it is allowed to add custom tags.

funcionante:
Thanks for the feedback. Now the unselection in the preset field should be correctly handled.

bastiao:
I will handle that.

bastiao:
Done.

PR #286, opened 2017-02-06
Add version in frontend maven plugin

bastiao:
- Added version in frontend maven plugin (to avoid unknown version warning)
- Added to the README the command to skip install npm and node.

PR #323, opened 2018-01-02
Add version to maven frontend plugin

Enet4:


Enet4:
Duplicate of #286, will update that one instead.

PR #326, opened 2018-02-01
Discard expired sessions and correctly handle login redirects

### LINKED ISSUE #219: Have the browser discard expired Dicoogle sessions
If the browser is holding an old or unexistent Dicoogle session, it should redirect the user back to the login view. By not doing so, services covered by access control (such as /webui) will simply fail without telling what's wrong.

funcionante:
Resolves #219. Depends on #301. Addresses these issues:

- [core] fix a bug in the code related to the endpoint `/login` (UsersXML.java) that would cause it to fail with an internal error (code 500) by retrieving roles in a user that were previously loaded as null values;
- [webapp] force login when the session is expired (verified on the server side);
- [webapp] after the login, let the website correctly redirect back to the page that was going to be visited before the login being forced (it was only working correctly when guest credentials were active);
- [webapp] avoid getting stuck on the loading page when accessing it;
- [webapp] when already logged in, avoid being redirected to the login page if accessing the root page.

Some changes were performed to prevent redundancies that caused unpredictable behavior. Please warn me if I excluded any important step during the refactoring.

### DIFF
--- a/dicoogle/src/main/resources/webapp/js/app.js
+++ b/dicoogle/src/main/resources/webapp/js/app.js
@@ -41,7 +41,8 @@ class App extends React.Component {
 		super(props);
 		this.needsPluginUpdate = true;
 		this.state = {
-			pluginMenuItems: []
+			pluginMenuItems: [],
+      lastLocation: 'search'
 		};
 		this.dicoogle = dicoogleClient(Endpoints.base);
 		this.logout = this.logout.bind(this);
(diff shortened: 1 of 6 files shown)


dicoogle/src/main/resources/webapp/js/app.js
```diff
@@ -64,33 +65,23 @@ class App extends React.Component {
 		});
 	}
 
+	componentWillMount() {
```
Enet4: There are a few mixtures of tabs and spaces in the code. Can you do a quick conversion to spaces (only on your changes)?

Enet4:
:house: Just rebased this branch

PR #328, opened 2018-02-08
Remove jQuery

### LINKED ISSUE #325: Remove jQuery

funcionante:
Resolves #325. The changes are the following:

- the jQuery datepicker in the advanced search section was replaced by [react-datepicker](https://github.com/Hacker0x01/react-datepicker); this new one allows traveling in the calendar, unlike the last one;
- the sidebar toggle actions in jQuery were replaced;
- the scrolling using jQuery in the logger view was replaced;
- the [DataTables](https://datatables.net/) (supported in jQuery) were removed in `directDumpView` (apparently they were not adding further functionality to the existing table);
- the jQuery transitions used in toasts were changed to be handled by [react-transition-group](https://github.com/reactjs/react-transition-group); if this change gets approved, I will apply it to #318 too, where toasts are also being used;
- the autocomplete in jQuery in the search view now uses [react-autosuggest](https://github.com/moroshko/react-autosuggest); before using [react-autosuggest](https://github.com/moroshko/react-autosuggest) I tried [react-select](https://github.com/JedWatson/react-select) (previously used in other changes), but I didn't found a way to adjust it to the scenario of choosing a suggestion and continuing typing, because [react-select](https://github.com/JedWatson/react-select) insists on resetting the input when a suggestion is chosen;
- The last change also introduced a fix to a bug where accepting a suggestion starting in the third term in a query (eg. `a AND b AND suggestion`) would result in a wrong edition in the input to (`ab AND suggestion`).

These changes replace almost all the jQuery usages in the app. The exceptions go to the HTTP requests, that were already replaced in #315; and to app.js, that apparently is referring jQuery to make bootstrap work (in the code, see _Bootstrap won't work without this hack. browserify-shim didn't help either_).

All the introduced components seem to be behaving correctly but for some of them, in the installation process, npm warms to the usage of later versions of `react` and `react-dom`. Maybe we should consider updating these components.


funcionante:
Thanks. I reverted the changes in the datepicker and kept only the text field, with no jQuery. These changes are now in the branch `wip/rework-advanced-search`.

Also, `jquery` was updated to `^3.2.1` and `jquery-ui` was removed. As far as I tested, the webapp keeps behaving as before.

Enet4:
Neat. I also noticed that the jquery-ui stylesheet is still included. Can you remove jquery-ui.css and the respective `<link>` element in our HTML template (line 31)? After this, it should be ready to merge.

Enet4:
I found something strange on Chromium 64.0.3282.140. I would like you to try to reproduce this:

1. Log in to Dicoogle.
2. Write a search query and press Enter. The browser navigates to the same page (like a page refresh was triggered) when it shouldn't.
3. Log out. The page also refreshes right after logging out, unnecessarily.

I couldn't reproduce this on Firefox 58.

funcionante:
Well noted, the same happened to me in Chromium. It seems to be related to the problem described [here](https://github.com/react-bootstrap/react-bootstrap/issues/1510). Please confirm if it is fixed now.

Also, I changed the search input to be focused when the view is rendered. Is it ok?

PR #329, opened 2018-02-08
Update README.md

eriksson:
- Update Dicoogle logo
- Update Monteiro's affiliation


Enet4:
The Dicoogle logo update is already made in #314. Maybe we could just update your affiliation over there and close this one?

PR #339, opened 2018-03-15
Document Web API

funcionante:
This introduces an OpenAPI specification for the server's web services. The file added can be previewed in Swagger Editor, [here](https://editor.swagger.io/).


Enet4:
Just for a little more context, this PR is the first step towards the creation of a Dicoogle technical reference hub. We intend to use the GitHub pages of the core repository to keep a full specification of both Web API and javadocs. The front page can be built manually, but I hope that we can generate the docs from the sources automatically with some kind of CI hook (something like [this](https://gist.github.com/willprice/e07efd73fb7f13f917ea)).

bastiao:
It seems a good idea, indeed.

@Enet4 anyway to check the final output of the technical reference? And regarding the build? Built manually could be complex and after a while it will be out of sync.

Enet4:
@bastiao Yes, you can open the [Swagger UI live demo](http://petstore.swagger.io) or the online Swagger editor and import the URL to the raw [dicoogle_web_api.yaml](https://raw.githubusercontent.com/bioinformatics-ua/dicoogle/947257074e1106a498233caea331a2c40bb3ee3a/dicoogle_web_api.yaml).
For generating the final output, this can either be done with [swagger-codegen](https://github.com/swagger-api/swagger-codegen/) or [spectacle](https://github.com/sourcey/spectacle). Both seem to work well here.

For integration purposes, we can start by making a script for generating the latest docs and pushing them to the GH pages. Later on, if rapid documentation updates are found to be important, it is possible to let a GitHub application do this automatically with some kind of commit hook.
The same idea can be applied to other technical docs (javadoc), which we do intend to include.

I'm also starting to think that there are more benefits from this file being in the `master` and `dev`  branches instead, because then they can be updated side-by-side with the related changes in the implementation.

bastiao:
@Enet4 indeed. I know Swagger of course, just wanted to know if you pushed automatically for somewhere or need to built manually. 

Nice to know, we hope that we can push it soon.

And also agree, keep docs in master/dev branches - definitely we should do it, and we can always keep nightly builds with latest updated docs.

Enet4:
Hmm, it seems that changing the base branch isn't enough, it has to be rebased. I'll send in a new PR.

Issue #122, opened 2015-09-23
Acknowledge which query providers list DICOM meta-data


Enet4:
Indexer plugins exist to serve two key purposes:
1. To make an abstraction over the underlying database technology, thus being able to benchmark different DICOM indexation techniques. Query providers bound to this indexer will have the necessary logic to retrieve DICOM data from such a database.
2. To obtain different interpretations over the DICOM data (e.g. applying CBIR, statistical analysis, ...). Query providers will allows us to retrieve the resulting data (e.g. feature vectors).

With that in mind, we know that some indices will hold DICOM data, while others will not. Only the query providers that hold DICOM should be selected for operations such as retrieving DIM identifiers (StudyInstanceUID, SeriesInstanceUID, SOPInstanceUID), and all DICOM-specific protocols (Q/R, WADO, ...). However, right now there's no way to know where DICOM data is kept. Even if some query providers may understand our Lucene-based queries, they might not give us the intended results, and our DICOM service implementations could (in the worst case) fail against this situation.

The actual request: To have a means in the Dicoogle core to identify which of the query providers are DICOM providers. A few possible approaches:
- _Have a list of DICOM providers in the global configuration file._ It requires manual intervention on installing a new plugin (this piece of information is fixed and plugin-dependant), but requires no modification to the SDK or existing plugins. **Added:** If no query provider was listed here, then it is assumed that all providers are DICOM.
- _Add a new method_ to `QueryInterface` with a signature similar to `bool providesDICOM()`, thus leaving the flag defined by the plugin itself, as it's meant to be. Unfortunately, this demands a change to the SDK, as well as existing plugins. 
- A mid-term: _Have a flag in the plugin's local configuration file._ The plugin can force it to a constant value and the core can read these files too. Only requires modifying the plugin controller and our DICOM providers.

Have any additional ideas? What do you think?

Enet4:
I will take the first approach while keeping the same behaviour if no specific configuration was given.

> Have a list of DICOM providers in the global configuration file. It requires manual intervention on installing a new plugin (this piece of information is fixed and plugin-dependant), but requires no modification to the SDK or existing plugins. If no query provider was listed here, then it is assumed that all providers are DICOM.

bastiao:
After the discussion, and agreement, I will only highlight  the points for the record:
if there are no defined settings, all the plugins will be assumed as DICOM.
Next major jump, we may think to adopt SDK changes.

Enet4:
[...]

PR #342, opened 2018-03-28
Use query providers that are DICOM

funcionante:
This targets the last update in #122. Two things that might be important to note:
- I used Java 8; since we are not supporting Java 7 anymore, I thought it would be ok;
- In `GET searchDIM/` I am ignoring providers that are not valid, instead of throwing an error; however, the web app is handling that, when using a multi-select that validates the providers; is it ok?

Also, possibly I forgot to apply the new settings somewhere. Please tell me if something is missing.


dicoogle/src/main/java/pt/ua/dicoogle/server/web/servlets/search/DumpServlet.java
```diff
@@ -58,11 +55,13 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
         }
-        
+
         String[] providerArr = req.getParameterValues("provider");
-        List<String> providers = (providerArr == null)
-                ? PluginController.getInstance().getQueryProvidersName(true)
-                : Arrays.asList(providerArr);
+        List<String> providers = PluginController.getInstance().filterDicomQueryProviders(providerArr);
+
+        if (providers.size() == 0) {
```
Enet4: I was initially thinking about only using the internal dimProvider list if the provider parameter was empty. But I suppose this works too. We'll just make sure that both Bioinformatics and BMD agree on this API.

dicoogle/src/main/java/pt/ua/dicoogle/server/web/servlets/search/DumpServlet.java
```diff
@@ -58,11 +55,13 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+            resp.sendError(400, "No DIM providers supplied.");
```
Enet4: This servlet could use the same fix as the one in `/search`, but I guess we can postpone this.

funcionante:
Thank you for the feedback. I applied the changes you proposed to `filterDicomQueryProviders()`.

bastiao:
@Enet4 there is only one dim provider supported in settings? We will need more than one. Any reason why only single plugin is supported at settings? 

There is also missing the DICOM Q/R part, but I can handle.

Enet4:
@bastiao Yes, it should work for multiple DIM providers. What you seem to have found is a bug in the server settings deserialization process, which should have admitted multiple `<dim-provider>` elements into the same list. I am currently looking into that.

bastiao:
@Enet4 yap, it should be

```
<dim-providers>
<dim-provider>a</dim-provider>
<dim-provider>b</dim-provider>
...
<dim-provider>n</dim-provider>
</dim-providers>
```

PR #344, opened 2018-04-11
Always use temporary directory when exporting to CSV

Enet4:
This is a recently discovered bug which could potentially affect everyone using the export to CSV functionality. The server's `Path` configuration was being used as the base folder for keeping temporary csv outputs (so that they can be subsequently downloaded), rather than a system's temporary folder. If the folder specified in `Path` did not exist, the operation would fail with an HTTP error code 500.


bastiao:
@Enet4 I believe the problem was already fixed. 

```
        File tempDir = Paths.get(System.getProperty("java.io.tmpdir")).toFile();
```

Still, if you want to rebase your branch to  "Files.createTempDirectory("dicoogle")", I also like the idea.

Enet4:
@bastiao I don't understand your request here. Where did you find that line of code?

This is to be pushed to `master` (and then to `dev`) because active users of Dicoogle 2.x can still benefit from this fix. There doesn't seem to be a reason to rebase either.

bastiao:
@Enet4 there was a fix already, pushed directly to dev (in other PR). 
But agree to propagate your fix to master and solve conflit in dev.

PR #354, opened 2018-05-23
Format the webapp with Prettier

Enet4:
I have configured the web application project to automatically format JavaScript and scss files using [Prettier](//prettier.io).

- `prettier` is the opinionated code formatting tool. It will ensure a uniform code style regardless of how it was originally written, allowing us to focus on important things.
- `husky` is a precommit hook manager for performing certain tasks before `git commit`.
- `lint-staged` performs certain tasks on staged files. I've configured it to run ESLint on JavaScript files, and Prettier on JS and SCSS files.

Running Prettier locally should be easy as well, with the use of `prettier` or the respective plugin for any supported IDE.

Consider checking commits 335b8e3 and d63e533 for a more manageable diff.


bastiao:
It's seems quite interesting to keep code formatting the same way. 
Is good to know that we could have exceptions: https://prettier.io/docs/en/ignore.html

PR #358, opened 2018-05-25
Fixed issue related with previous change in QueryDimInterface.

bastiao:
Fixed issue related with previous change in QueryDimInterface. The query interface was not able to reach the correct interfaces at plugins.

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
@@ -483,7 +483,7 @@ public class PluginController{
         List<String> baseProviders = ServerSettingsManager.getSettings()
                 .getArchiveSettings().getDIMProviders();
 
-        if (baseProviders.isEmpty()) {
+        if (baseProviders==null || baseProviders.isEmpty()) {
             baseProviders = PluginController.getInstance().getQueryProvidersName(true);
         }
 
@@ -533,7 +533,7 @@ public class PluginController{
 
     public JointQueryTask queryAll(JointQueryTask holder, final String query, final Object ... parameters)
     {
-        return queryAll(holder, query, DimLevel.INSTANCE, parameters);
+        return queryAll(holder, query, parameters);
     }
 
     public JointQueryTask queryAll(JointQueryTask holder, final String query, final DimLevel level, final Object ... parameters)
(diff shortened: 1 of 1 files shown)


dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
```diff
@@ -603,8 +622,35 @@ public JointQueryTask query(JointQueryTask holder, List<String> querySources, fi
...
+        //returns a tasks that runs the query from the selected query engine
+        String uid = UUID.randomUUID().toString();
+        Task<Iterable<SearchResult>> queryTask = new Task<>(uid, querySource,
+                new Callable<Iterable<SearchResult>>(){
+                    @Override public Iterable<SearchResult> call() throws Exception {
+                        if(queryEngine == null || !(queryEngine instanceof QueryDimInterface)) return Collections.emptyList();
+                        try {
+                            return queryEngine.query(query, level, parameters);
+                        } catch (RuntimeException ex) {
+                            logger.warn("Query plugin {} failed unexpectedly", querySource, ex);
```
Enet4: A previously unattended issue: should public `query` methods in `DicooglePlatform` swallow the query-related exceptions `QueryException` and `RetrievalException`? Can we make them appear when attempting to get the outcome of the task?

dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
```diff
@@ -483,7 +483,7 @@ public QueryInterface getQueryProviderByName(String name, boolean onlyEnabled){
         List<String> baseProviders = ServerSettingsManager.getSettings()
                 .getArchiveSettings().getDIMProviders();
 
-        if (baseProviders.isEmpty()) {
+        if (baseProviders==null || baseProviders.isEmpty()) {
```
Enet4: Good catch. I had the feeling that this wasn't working properly in my recent uses of Dicoogle 3.

dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
```diff
@@ -603,8 +622,35 @@ public JointQueryTask query(JointQueryTask holder, List<String> querySources, fi
...
+    private Task<Iterable<SearchResult>> getTaskForQueryDim(final String querySource, final String query,
+                                                         final DimLevel level, final Object ... parameters){
+
+        final QueryInterface queryEngine = getQueryProviderByName(querySource, true);
+        //returns a tasks that runs the query from the selected query engine
+        String uid = UUID.randomUUID().toString();
+        Task<Iterable<SearchResult>> queryTask = new Task<>(uid, querySource,
+                new Callable<Iterable<SearchResult>>(){
+                    @Override public Iterable<SearchResult> call() throws Exception {
+                        if(queryEngine == null || !(queryEngine instanceof QueryDimInterface)) return Collections.emptyList();
```
Enet4: A more fail-fast approach would be performing this check before the task is created, which should be fine if you're OK with throwing when the program fails to create the task.

dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
bastiao: LGTM!

dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
bastiao: Yeap!

dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
bastiao: This is a private method - safer and comfortable to return an empty list. No clear benefit to throw exception IMHO and throw error at this stage.

PR #359, opened 2018-05-28
 Fix problem related with set SOP Class UID during bootstrap of DicomStorage

bastiao:
Fix problem related with set SOP Class UID during bootstrap of DicomStorage (C-STORE-SCP)
Note: This is a short and convenient fix. SOPClasses usage over DicomStorage needs a refactor anyway.

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/server/SOPList.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/server/SOPList.java
@@ -209,7 +209,14 @@ public class SOPList {
             TS.setAccepted(a);
         }        
         return 0;    
-    }   
+    }
+
+
+    public int updateTSFieldByTsUID(String sopUID, String tsUID, boolean value){
+        return updateTSField(sopUID, TransfersStorage.globalTransferUIDsMap.get(tsUID), value);
+    }
+
+
     /**
      * Updates a given SOP Class accepted Tranfer Syntaxes
      * @param UID SOP Class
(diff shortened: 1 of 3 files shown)


dicoogle/src/main/java/pt/ua/dicoogle/server/SOPList.java
```diff
@@ -218,7 +225,7 @@ public synchronized int updateTS(String UID, boolean [] p, boolean a) {
      * @return -1 if something went wrong, 1 otherwise
      */
     public synchronized int updateTSField(String UID, String name, boolean value) {
-        logger.debug("UID: {}, name: {}, value: {}", UID, name, value);
+        logger.info("UID: {}, name: {}, value: {}", UID, name, value);
```
Enet4: Given that this method is being called inside a 2-level loop, logging at info level is going to be quite spammy. Is this intentional?

dicoogle/src/main/java/pt/ua/dicoogle/server/TransfersStorage.java
```diff
@@ -47,6 +47,7 @@
      */
     
     public static final Map<Integer, String> globalTransferMap;
+    public static final Map<String, String> globalTransferUIDsMap;
```
Enet4: Odd, I would expect this dictionary to already exist in dcm4che... Would [`UIDDictionary`](https://github.com/dicomgrid/dcm4che2/blob/master/dcm4che-core/src/main/java/org/dcm4che2/data/UIDDictionary.java) work here?

bastiao: No. This dictionary does not work in the way that I need. It returns a different UID description.

Issue #193, opened 2016-01-30
Correct web service error reporting


Enet4:
Many of our web services, although normally using JSON, still report errors in HTML:

![screenshot_2016-01-30_15-25-35](https://cloud.githubusercontent.com/assets/4738426/12696576/49c09212-c766-11e5-9ddd-7afaa51d367b.png)

We should normalize these cases, providing error messages of the same content type when possible. This is important for the detection of bugs and errors on the clients.

Enet4:
I think we already did this for the most important services, but it's something that we should adapt organically as we touch them for other reasons.

PR #360, opened 2018-05-30
Validate UUID provided to /exportFile

Enet4:
This checks whether the identifier to the exported CSV file is actually a UUID. It fixes a vulnerability on Windows systems which would allow a user to retrieve an arbitrary file from the server.

I took the liberty of adapting this endpoint to our JSON error output (#193), and ensuring that the content-type is JSON when sending an error with our utility function.


bastiao:
@Enet4 Good. But in which part of the diff is located the "check whether the identifier to the exported CSV file is actually a UUID."?

Thanks and credits for the reporter.

Enet4:
@bastiao It's over [here](https://github.com/bioinformatics-ua/dicoogle/pull/360/files#diff-5fc326595cbb0b1fc3f38c64591d49edR67). The static method throws an exception if the string is not a valid UUID.

bastiao:
Got it. Sure, I will test in Windows Server, but LGTM.

PR #363, opened 2018-06-15
Docker support for dicoogle

organom:
Simple Dockerfile to allow the creation of a docker image for dicoogle. 

The docker image is based in the latest ubuntu image and will allow both build and run of dicoogle from inside the container.
 - Build :   docker build -t bioinformatics-ua/dicoogle .
 - Run :    docker run -t bioinformatics-ua/dicoogle
 - Access container :    docker run -it bioinformatics-ua/dicoogle bash

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,15 @@
+FROM ubuntu
+
+WORKDIR /root
+
+RUN apt-get update && \
+    apt-get install apt-utils -y && \
+    apt-get upgrade -y
+
+RUN apt-get install -y software-properties-common build-essential openjdk-8-jre git maven curl
+RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - && apt-get install -y nodejs
+
+RUN git clone https://github.com/bioinformatics-ua/dicoogle
+RUN ( cd dicoogle && mvn install && ln -s /root/dicoogle/dicoogle/target/dicoogle.jar /root/ )
+
+CMD ["java","-jar","dicoogle.jar","-s"]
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+RUN apt-get update && \
+    apt-get install apt-utils -y && \
+    apt-get upgrade -y
+
+RUN apt-get install -y software-properties-common build-essential openjdk-8-jre git maven curl
+RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - && apt-get install -y nodejs
+
+RUN git clone https://github.com/bioinformatics-ua/dicoogle
+RUN ( cd dicoogle && mvn install && ln -s /root/dicoogle/dicoogle/target/dicoogle.jar /root/ )
```
Enet4: The way pom.xml is currently set up, Maven will automatically fetch and locally install the LTS version of Node.js and npm for building the webapp. If you would like to use the one already installed, consider modifying this line to something like this:

```Dockerfile
RUN ( cd dicoogle/dicoogle/src/main/resources/webapp && npm install && cd ../../../../.. && mvn install -Dskip.installnodenpm -Dskip.npm && ln -s /root/dicoogle/dicoogle/target/dicoogle.jar /root/)
```

organom: You're right, the pom already gets it. 
So i think the best and easier way to do this is to remove the line of the nodejs installation and just let the mvn install everything. This way you can control version of npm and nodejs from the pom without having to think about the docker.

bastiao: @organom LGTM!

PR #364, opened 2018-07-09
Imp/webapp style

rlebre:
Updated webapp's color scheme:

screen shot 2018-07-09 at 17 58 03
screen shot 2018-07-09 at 17 57 49
screen shot 2018-07-09 at 17 57 40
screen shot 2018-07-09 at 17 55 22

Also updated basic styles.
**Before:**
screen shot 2018-07-09 at 18 06 57
screen shot 2018-07-09 at 18 06 49

**Now:**
screen shot 2018-07-09 at 18 04 59
screen shot 2018-07-09 at 18 04 49


dicoogle/src/main/resources/webapp/sass/modules/_management.scss
```diff
@@ -31,6 +31,7 @@
       color: $blue1;
       &:hover {
         background: $hover_color;
+        color: #ffffff;
```
bastiao: can we also use variable here and also in sequel? :)

Butt very nice piece of work! :)

rlebre: We can indeed. Last commits replace that case and other cases of button color with a variable :)

rlebre:
screen shot 2018-07-10 at 16 11 40
screen shot 2018-07-10 at 16 11 07

bastiao:
LGTM. 👍 
Very nice initiative!

Issue #223, opened 2016-05-04
Adjust user authentication for compliance


Enet4:
At the moment, consumers of our web services should provide an `Authorization` header on their request, containing nothing more but the Dicoogle session token. According to [RFC2617](http://www.ietf.org/rfc/rfc2617.txt), this header should actually contain credentials of the form below, for compliance:

>    credentials = auth-scheme #auth-param

Therefore, we should either use this header properly, keep this information in another header, or rely on other mechanisms for this purpose.

rlebre:
Let me refer to the [RFC 7235](https://tools.ietf.org/html/rfc7235) as it obsoletes and updates the RFC 2617.

Also, once we are discussing this matter, we should also address the JWT [RFC 7519](https://www.rfc-editor.org/info/rfc7519).

PR #388, opened 2019-07-10
Revamp user file handling and cryptographic code

### LINKED ISSUE #387: Misuses of cryptographic APIs
Hi
The following lines have cryptographic API misuses. <br> File name => dicoogle/src/main/java/pt/ua/dicoogle/server/users/UserFileHandle.java: Line number => 84: API name => Cipher:Operation on object of type javax.crypto.Cipher object not completed. Expected call to init<BR> File name => dicoogle/src/main/java/pt/ua/dicoogle/server/users/UserFileHandle.java: Line number => 141: API name => Cipher:<BR> File name => dicoogle/src/main/java/pt/ua/dicoogle/server/users/HashService.java: Line number => 45: API name => MessageDigest:<BR> File name => dicoogle/src/main/java/pt/ua/dicoogle/server/users/UserFileHandle.java: Line number => 141: API name => Cipher:<BR> File name => dicoogle/src/main/java/pt/ua/dicoogle/server/users/HashService.java: Line number => 43: API name => MessageDigest:First parameter (with value "SHA-1") should be any of {SHA-256, SHA-384, SHA-512}<BR> File name => dicoogle/src/main/java/pt/ua/dicoogle/server/users/UserFileHandle.java: Line number => 103: API name => Cipher:<BR> File name => dicoogle/src/main/java/pt/ua/dicoogle/server/users/UserFileHandle.java: Line number => 103: API name => Cipher:<BR>

Enet4:
I believe it fully resolves #387.

- Replace all uses of the SHA-1 algorithm with BCrypt, a more suitable password hashing algorithm. It's still worth noting that this codebase was using SHA-1 far before the SHAttered attack was known.
- Make user file encryption actually work. This will happen by default unless a users file already exists without the respective key file.
- Try to set owner read-only permissions to users encryption key
- Refactor code around the `Cipher` API. I'm still not quite convinced that `Cipher` was being misused, though.
- Change user credentials file name to just "users" (since it might be encrypted, the .xml extension is misleading)

This is another breaking change for Dicoogle, each new deployment will have to create user credentials again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ sudo: true
 
 before_cache:
  - rm -rf $HOME/.m2/repository/pt/ua/ieeta/dicoogle*
+ - rm -rf $HOME/.cache/JNA/temp/
 
 cache:
   directories:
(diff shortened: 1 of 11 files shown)


Enet4:
Pending decision: if we are going to replace the cryptographic hash algorithm for the user's passwords, we might as well use a more suitable algorithm (e.g. BCrypt).

rlebre:
The usage of a more suitable algorithm like BCrypt may lead us to another discussion in the meantime: we should support JWT instead of a simple session token.

Enet4:
@rlebre We have a more suitable issue for that, #223. I would be grateful if you provided your perspective there.

Enet4:
Status update of this PR: I replaced the current hashing algorithm with Argon2, a state-of-the-art password hashing algorithm. The bindings targeting the JVM for access to the native implementation (through JNA) were readily available with pre-compiled binaries via Maven.

However, the tests on Travis fail because the JNA access to argon2 requires a more recent version of glib than the one provided by Ubuntu in the trusty distribution. In turn, we are unable to update the distribution to a more recent version because those no longer support the installation of a JDK 8, which we haven't planned to drop support of just yet.

Although I know this works, the cost of keeping this implementation is too high. I'll look for another package or hashing algorithm that does not have the same problem.

PR #397, opened 2019-10-16
added option to disable web app

wesdean:
I have added a config option to disable the default web services. The reason for doing this is that the web services are not secure. This way I can use plugins to expose only the web services that I create with custom authentication. This feature may be useful to others.


bastiao:
Thanks for contribution, and indeed sorry for the late reply.
Currently, the code is not rebased with mainstream. Do you have still some interest in bring this change to the mainstream?

wesdean:
Unfortunately I do not have the time to work on this at the moment. This work was for a project that I have moved away from.

bastiao:
> Unfortunately I do not have the time to work on this at the moment. This work was for a project that I have moved away from.

Ok, no problem, thanks for your reply.

PR #404, opened 2019-12-17
Bug/webapp/plugins services

### LINKED ISSUE #379: Installed Plugins list not showing
The new form containing the list of installed plugins doesn't seem to be working in our latest development branch anymore. It just stays in the loading state.

ae08f955-8e65-4897-b8c6-c86e6bc85ff9

This feature was introduced in #305 (requested by #263). It might be worth tracing back through the commit history to see what led to this problem.

rlebre:
Fixes #379 

The DicoogleClient API was updated. Thus, the request in the WebApp got outdated. This PR fixes the issues in Management > Services and Plugins > Plugins.

### DIFF
--- a/dicoogle/src/main/resources/webapp/js/stores/pluginStore.js
+++ b/dicoogle/src/main/resources/webapp/js/stores/pluginStore.js
@@ -2,40 +2,40 @@ import Reflux from "reflux";
 import { Endpoints } from "../constants/endpoints";
 import * as PluginActions from "../actions/pluginActions";
 
-//import {request} from '../handlers/requestHandler';
-
 import dicoogleClient from "dicoogle-client";
 const Dicoogle = dicoogleClient(Endpoints.base);
 
 const PluginStore = Reflux.createStore({
   listenables: PluginActions,
-  init: function() {
+  init: function () {
     this._contents = {};
   },
 
-  onGet: function(type) {
+  onGet: function (type) {
     const self = this;
 
-    Dicoogle.request("GET", ["plugins", type], (error, data) => {
+    Dicoogle.request("GET", ["plugins", type]).end(function (error, data) {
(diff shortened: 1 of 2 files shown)


dicoogle/src/main/resources/webapp/js/stores/pluginStore.js
```diff
@@ -2,40 +2,40 @@ import Reflux from "reflux";
...
+  init: function () {
     this._contents = {};
   },
 
-  onGet: function(type) {
+  onGet: function (type) {
     const self = this;
 
-    Dicoogle.request("GET", ["plugins", type], (error, data) => {
+    Dicoogle.request("GET", ["plugins", type]).end(function (error, data) {
```
Enet4: You can keep using the arrow function syntax, so you can use `this` without having to create an outer `self`.

dicoogle/src/main/resources/webapp/js/stores/pluginStore.js
```diff
@@ -2,40 +2,40 @@ import Reflux from "reflux";
...
       }
     });
   },
 
-  onSetAction: function(type, name, action) {
+  onSetAction: function (type, name, action) {
     const self = this;
 
-    Dicoogle.request("POST", ["plugins", type, name, action], (error, data) => {
+    Dicoogle.request("POST", ["plugins", type, name, action]).end(function (error, data) {
```
Enet4: Same here.

rlebre:
Indeed! 
Thank you for your suggestions. In fact, that way we may not use the outer self. I fixed that and also fixed the warning 
`Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of 'PluginsView'. See https://fb.me/react-warning-keys for more information.`
when opening the Services and Plugins tab.

PR #420, opened 2020-05-17
Update react-bootstrap-table to v3

Enet4:
This PR updates the dependency `react-bootstrap-table` to the latest v3: 3.5.1.

This update is relevant because it reduces the total JavaScript bundle size from 1.5MB to 1.14MB, as a consequence of now being able to minify the dist code of this dependency properly. The production bundle constructed via  Webpack no longer contains calls to `eval`.

- update dependency + css file
- instruct webpack to resolve `global` to `window`
- replace deprecated use of `ReactBootstrapTable.Input`
- update `SearchResult` prop types
- remove statement of deprecation of `SearchResultView` (since it's not true)

- in progress: let the user go forward in the result steps if a previous selection at the same level had been made


dicoogle/src/main/resources/webapp/js/components/search/searchResult.jsx
```diff
@@ -326,8 +327,10 @@ const Step = React.createClass({
     else if (step > current) return "col-xs-3 wizardbar-item disabled";
     else if (step < current) return "col-xs-3 wizardbar-item completed";
   },
-  onStepClicked: function(current) {
-    this.props.onClick(current);
+  onStepClicked: function(newStep) {
+    if (this.props.current > newStep) {
```
Enet4, line 331: There's a regression here: in the previous version, you could enter a more specific step (e.g. Series) if you had already selected something at that level.

An example workflow:

1. Select a patient,
2. Select a study,
3. Go back to the patient list,
4. Select the Study step.

This would still work, but not anymore with these changes.

bastiao:
Ok, @Enet4 let me know what you would like to do it here, so we can move forward with this.

Enet4:
I would be interested in resolving the regression, so that one can go back and forth in the DICOM tree more freely.

I'll mark the PR as ready when I'm done.

bastiao:
Hi. It does not seems to appear in a good way, after merge 'imp/update-react-bootstrap-table-3' into dev. Thoughts? 

image

Enet4:
@bastiao This is what I reproduce on a 1080p screen in Chrome 86. Could you please expand how you reproduced that? This branch is already based on `dev`. Some effort was already made for the table columns to have a reasonable presentation quality even under more restrictive window dimensions.

Screnshot-2020-12-23_dicoogle_search

PR #435, opened 2020-12-11
Update Travis badge in README

Enet4:
As required by the migration from travis-ci.org to travis-ci.com.


bastiao:
Failed with WEB_APP only, any idea why?

Enet4:
> Failed with WEB_APP only, any idea why?

I strongly suspect that the current node_sass no longer compiles with Node.js 15. We can try to update this one or look for alternatives in plain JS.

bastiao:
> > Failed with WEB_APP only, any idea why?
> 
> I strongly suspect that the current node_sass no longer compiles with Node.js 15. We can try to update this dependency or look for alternatives in plain JS.
> 
> This shouldn't be a problem with the rest of the project, so long as we stick to supported versions up to LTS.

Yes, we can, but this could be a separate task to avoid block the release. If it currently compile in Node 14 which is LTS, it should be fine for now.

PR #436, opened 2020-12-11
Update sass dev dependencies

Enet4:
Fixes build errors using latest stable (15), while still staying within our supported Node.js versions (10, 12, 14, 15).

Also replaces `node-sass`, with `dart-sass`, which is recommended by sass-loader and doesn't require a compilation step.


Enet4:
This should be good to merge after merging #437.

PR #440, opened 2020-12-17
Fix a bug while deleting a study with too many images

Rui-Jesus:
This change increases the default maxFormContentSize to allow operations such as unindex/delete over large studies (with >3k image).


bastiao:
I'm changing the base. I will propagate to dev branches manually due to intention of keep a minor release in 2.5.X branch.

Issue #146, opened 2015-10-17
Support Dicoogle task registering at plugin-level.


Enet4:
At the moment, Dicoogle tasks are only used for indexation purposes. I propose that plugins should be able to register their own long-running tasks (e.g heavy data analysis, vocabulary generation, etc.) so that they can be listed and terminated in our services.

PR #441, opened 2020-12-22
Fix memory leak in the completed/failed indexing task.

### LINKED ISSUE #139: [webapp] high number of tasks breaks the index page
Some possibles to fix it:

- Pagination
- Truncate number of tasks to show
- Only shows some of the tasks in this interface.

bastiao:
This PR avoids to keep all indexing tasks in memory.
The oldest not running tasks are cleaned, to guarantee a hooked memory cleaning process. 

Base is master, but planned to merge in dev as well.

Resolves #139.

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/taskManager/RunningIndexTasks.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/taskManager/RunningIndexTasks.java
@@ -40,7 +40,7 @@ public class RunningIndexTasks {
     private static final Logger logger = LoggerFactory.getLogger(RunningIndexTasks.class);
     private static Integer SOFT_MAX_RUNNINGTASKS = Integer.parseInt(System.getProperty("dicoogle.tasks.softRemoveTasks", "50000"));
     private static Integer NUMBER_RUNNINGTASKS_TO_CLEAN = Integer.parseInt(System.getProperty("dicoogle.tasks.numberTaskClean", "2000"));
-    private static Boolean ENABLE_HOOK = Boolean.getBoolean(System.getProperty("dicoogle.tasks.removedCompleted", "true"));
+    private static Boolean ENABLE_HOOK = Boolean.valueOf(System.getProperty("dicoogle.tasks.removedCompleted", "true"));
 	public static RunningIndexTasks instance;
 
 	private final Map<String, Task<Report>> taskRunningList;
--- a/dicoogle/src/main/java/pt/ua/dicoogle/taskManager/RunningIndexTasks.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/taskManager/RunningIndexTasks.java
@@ -18,14 +18,6 @@
  */
 package pt.ua.dicoogle.taskManager;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


dicoogle/src/main/java/pt/ua/dicoogle/taskManager/RunningIndexTasks.java
```diff
@@ -52,17 +54,38 @@ public static RunningIndexTasks getInstance() {
...
 	}
 
+
+	public boolean hookRemoveRunningTasks(){
+	    if (this.taskRunningList.size()>SOFT_MAX_RUNNINGTASKS){
+	        int removedTasks = 0 ;
+	        Iterator<String> iterator = this.taskRunningList.keySet().iterator();
+	        while(iterator.hasNext()&& removedTasks<NUMBER_RUNNINGTASKS_TO_CLEAN){
+	            String tId = iterator.next();
+                Task t =  this.taskRunningList.get(tId);
```
Enet4, line 79: Raw type.

```suggestion
                Task<?> t =  this.taskRunningList.get(tId);
```

dicoogle/src/main/java/pt/ua/dicoogle/taskManager/RunningIndexTasks.java
```diff
@@ -52,17 +54,38 @@ public static RunningIndexTasks getInstance() {
...
+                if (t.isCancelled() || t.isDone()){
+                    iterator.remove();
+                    removedTasks++;
+                }
+            }
+        }
+	    return true;
```
Enet4, line 86: The method always returns `true`. It should either provide additional information through the returned value (e.g. whether at least one task was removed) or just return `void`.

Enet4 (commented):
The concrete effect of these changes are that, upon the addition of a new task, `N` cancelled or stopped tasks are synchronously deleted from the list of tasks whenever we have more than `M` of them. `(N, M)` are `(2_000, 50_000)` by default.

We can move on with this proposal, as the issues arriving from a large number of task descriptors sitting in memory need to be resolved. Still, interest should be kept towards the resolution of #146 and #302. 

- #146 An integrated and generic task execution service, with centralised awareness of the process allocated resources, instead of relying on independent task pools per plugin.
- #302 A full, searchable history of tasks, enabling better task auditing without custom logging configurations.

Also, minor code improvement suggestions follow.

dicoogle/src/main/java/pt/ua/dicoogle/taskManager/RunningIndexTasks.java
bastiao: Thanks for feedback @Enet4. I will address the minor issues. The code at `dev` branch is already formated by the plugin already, so it will be easier to merge.

Regarding the other issues, sure, we can kept them for newest versions of Dicoogle 3 or 4 . For now, we need to close this one, close the 3.0.0 and move forward :)

Enet4: > The code at dev branch is already formated by the plugin already, so it will be easier to merge.

Ah right, the 2.X development branch does not have standardized code formatting. We can look into that later if it becomes a nuisance. :)

I say it's good to merge once the other minor concerns are resolved.

PR #442, opened 2021-01-03
Update webapp to React v15

Enet4:
This PR updates multiple dependencies of the web application so that it can use React 15. This enables debugging and profiling React applications via the latest devtools, as well as for an easier path towards updating the project to the latest version of React, v17, which offers greater levels of flexibility in terms of what components we can use.

A summary:

- Update `react`, `react-dom` and `react-addons-update` to 15
- Update `reflux` to v5
- Update `react-bootstrap` to 0.32
   - version v1 is already available, but it requires React v16
- Replace dependency `react-imageloader` (no longer maintained) with `react-load-image`
- Remove attempt at using non-existing asset spinner.gif on image load, use loader instead
- Rename webcore/test to webcore/examples
- Update some of the web plugin examples with the new generator-based structure
- Replace all uses of `React.PropTypes` with `PropTypes` from package 'prop-types'
- Replace all uses of `React.createClass` with either classes extending `React.Component` (recommended) or to use createReactClass from 'create-react-class'
- Add check on user store for whether Dicoogle had been authenticated before checking user login, prevents superfluous requests returning HTTP 401
- react-router was kept at 3, _not_ updated to version 4, because v4 is a heavy rewrite with a substantially different API. v3 still supports React from 0.14 to 16.


bastiao:
It seems to work fine. The WebPlugins seems to be broken (but they were already)

PR #447, opened 2021-02-07
Small improvements in button UI for search result slot.

bastiao:
Small improvements in button UI for search result slot.

Before:

Screen Shot 2021-02-07 at 17 22 06

After: 

slots-image

The plugins that interects with search results are currently totally broken, but hopefully can submit a patch soon.


Enet4 (approved):
Yes, it is indeed better. :+1: I probably didn't leave an icon in the first iteration because the info that it's a web plugin isn't very interesting to have, and eventually as more plugins were in, we'd just have a bunch of plugs. :) But I guess it's better than nothing.

PR #448, opened 2021-02-07
WebUI Plugins: Fix the emit events for result-batch slot plugins

bastiao:
Currently, the result-batch plugins are not able to receive the list of (patients,studies, series, images) to apply actions. 
Actually it fails before to try to emit the proper values. There are two fixes:
- Webcore: allow to emit signals 
- Patient,Study.. Images: properly get a list of images, was probably failing due to dependency updates such as react, react-bootstrap and all these related matters.

### DIFF
--- a/dicoogle/src/main/resources/webapp/js/components/plugin/pluginForm.jsx
+++ b/dicoogle/src/main/resources/webapp/js/components/plugin/pluginForm.jsx
@@ -1,8 +1,7 @@
 import React from "react";
 import * as PropTypes from "prop-types";
 import { ResultsSelected } from "../../stores/resultSelected";
-import dicoogleClient from "dicoogle-client";
-const Dicoogle = dicoogleClient();
+import Webcore from "dicoogle-webcore";
 
 export default class PluginForm extends React.Component {
   static get propTypes() {
@@ -38,7 +37,8 @@ export default class PluginForm extends React.Component {
       const node = component;
       node.data = this.props.data;
       node.addEventListener("hide", this.handleHideSignal);
-      Dicoogle.emitSlotSignal(
+
+      Webcore.emitSlotSignal(
         node,
         "result-selection-ready",
         ResultsSelected.get()
--- a/dicoogle/src/main/resources/webapp/js/components/search/result/imageView.js
+++ b/dicoogle/src/main/resources/webapp/js/components/search/result/imageView.js
@@ -154,12 +154,9 @@ const ImageView = createReactClass({
   handleSelect(item) {
     let { sopInstanceUID } = item;
     // ResultSelectActions.select(item);
-    let value = this.refsClone[sopInstanceUID].getChecked();
+    let value = this.refsClone[sopInstanceUID].checked;
     if (value) ResultSelectActions.select(item, sopInstanceUID);
     else ResultSelectActions.unSelect(item, sopInstanceUID);
-  },
-  handleRefs: function(id, input) {
-    this.refsClone[id] = input;
   },
   formatSelect: function(cell, item) {
     let { sopInstanceUID } = item;
@@ -169,7 +166,7 @@ const ImageView = createReactClass({
         <Checkbox
           label=""
           onChange={this.handleSelect.bind(this, item)}
-          ref={this.handleRefs.bind(this, sopInstanceUID)}
+          inputRef={ref => this.refsClone[sopInstanceUID] = ref}
         />
       </div>
     );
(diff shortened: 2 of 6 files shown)


dicoogle/src/main/resources/webapp/js/components/plugin/pluginForm.jsx
```diff
@@ -1,8 +1,7 @@
 import React from "react";
 import * as PropTypes from "prop-types";
 import { ResultsSelected } from "../../stores/resultSelected";
-import dicoogleClient from "dicoogle-client";
-const Dicoogle = dicoogleClient();
+import Webcore from "dicoogle-webcore";
```
Enet4, line 4: :+1: Injecting methods into the global Dicoogle access object wasn't that good of an idea after all, so I'm OK with this.

dicoogle/src/main/resources/webapp/js/components/search/result/patientView.js
```diff
@@ -138,7 +135,7 @@ const PatientView = createReactClass({
         <Checkbox
           label=""
           onChange={this.handleSelect.bind(this, item)}
-          ref={this.handleRefs.bind(this, id)}
+          inputRef={ref => this.refsClone[id] = ref}
```
Enet4, line 138: This `handleRefs` as a separate function was intentional, so that it did not have to create an arrow function on each render call (more info [here](https://reactjs.org/docs/handling-events.html)). If `ref` instead of `inputRef` was the real problem, then `handleRefs` could be brought back.

bastiao: Yes, I agree. There is one more mistake at patient level, going to fix it.

dicoogle/src/main/resources/webapp/js/components/search/result/patientView.js
```diff
@@ -122,11 +122,14 @@ const PatientView = createReactClass({
   },
   handleSelect(item) {
     let { id } = item;
-    ResultSelectActions.select(item);
+    // ResultSelectActions.select(item);
```
Enet4, line 125:
```suggestion
```

dicoogle/src/main/resources/webapp/js/components/search/result/imageView.js
```diff
@@ -169,7 +168,7 @@ const ImageView = createReactClass({
         <Checkbox
           label=""
           onChange={this.handleSelect.bind(this, item)}
-          ref={this.handleRefs.bind(this, sopInstanceUID)}
+          inputRef={this.handleRefs.bind(this, sopInstanceUID)}
```
Enet4, line 171: Don't forget the other components.

dicoogle/src/main/resources/webapp/js/components/search/result/studyView.js
```diff
@@ -113,12 +113,12 @@ const StudyView = createReactClass({
...
     // ResultSelectActions.select(item);
-    let value = this.refsClone[studyInstanceUID].getChecked();
+    let value = this.refsClone[studyInstanceUID].checked;
     if (value) ResultSelectActions.select(item, studyInstanceUID);
     else ResultSelectActions.unSelect(item, studyInstanceUID);
   },
-  handleRefs: function(id, input) {
-    this.refsClone[id] = input;
+  handleRefs: function(id, input) {	
+    this.refsClone[id] = input;
```
Enet4, line 121: Odd. Is this trailing whitespace? I thought prettier was in place to fix it automatically.

bastiao: Me too. It's heavy to have those pre commit hooks, so at least they can make some magic for us. It does not seems to be the case, here.

bastiao (commented):
> 👍 Injecting methods into the global Dicoogle access object wasn't that good of an idea after all, so I'm OK with this.

Ok, but I'm still not happy with that solution. Do you have a solution to emit signals from the plugin side? I was thinking to make the method also available in `HTMLDicoogleSlotElement`. Any suggestions? Is there a easy way to get `Webcore.emitSlotSignal` inside plugin?

bastiao:
> > 👍 Injecting methods into the global Dicoogle access object wasn't that good of an idea after all, so I'm OK with this.
> 
> Ok, but I'm still not happy with that solution. Do you have a solution to emit signals from the plugin side? I was thinking to make the method also available in `HTMLDicoogleSlotElement`. Any suggestions? Is there a easy way to get `Webcore.emitSlotSignal` inside plugin?

After offline discussion, `import Webcore from "dicoogle-webcore"` will do the job.

Enet4 (approved):

PR #450, opened 2021-02-16
Fix exportToCSV modal on webapp

rlebre:
This PR fixes the Export to CSV modal issues.
The button styles were changed and now match the web app style. Mobile version is considered.
The input field to name the export preset was also fixed and the user can now change the name.

Before:
Screenshot 2021-02-16 at 12 41 48

Now:
now

### DIFF
--- a/dicoogle/src/main/resources/webapp/sass/components/export.scss
+++ b/dicoogle/src/main/resources/webapp/sass/components/export.scss
@@ -2,3 +2,16 @@
   width: 100%;
   height: 400px;
 }
+
+.modal-dicoogle {
+  text-align: unset;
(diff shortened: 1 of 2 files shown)


dicoogle/src/main/resources/webapp/sass/components/export.scss
```diff
@@ -2,3 +2,16 @@
   width: 100%;
   height: 400px;
 }
+
+.modal-dicoogle {
```
Enet4, line 6: Not an exciting class name, but this does not worry me much.

Enet4 (approved):
Looks OK. Preset saving and loading works well alongside #451.

PR #452, opened 2021-02-22
Improve logging of C-MOVE SCP to better troubleshoot potential issues.

bastiao:
Improve logging of C-MOVE SCP to better troubleshoot potential issues.
A different patch may need to be introduced for dev branch.

PR #453, opened 2021-02-27
DICOM services - bring patches from 2.5.X releases to main stream 

bastiao:
Bringing two patches to the mainstream:

- Applying the allow to change DICOM C-STORE SCP parameters #446
- Improve logging of C-MOVE SCP to better troubleshoot potential issues. #452


dicoogle/src/main/java/pt/ua/dicoogle/server/queryretrieve/CallDCMSend.java
```diff
@@ -36,6 +38,8 @@
  * @author Luís A. Bastião Silva <bastiao@ua.pt>
  */
 public class CallDCMSend {
+    private static final Logger logger = LoggerFactory.getLogger(CMoveServiceSCP.class);
```
Enet4, line 41: I don't know if this is the equivalent change from before, but the class names do not match. `CMoveSCP` instead of `CallDCMSend`.

Enet4 (approved):

PR #463, opened 2021-05-14
Update CI workflows to check Java and webapp separately

Enet4:
- add node.yml
   - build webapp using multiple versions of Node.js
- update maven.yml
   - build Java project using JDK 8 and 11
   - skip webapp
- update readme
   - some things do not have to be installed in the system


Enet4:
The failure on Node.js 16 is a known breaking change regarding how peer dependencies are handled, we can fix it later.

PR #482, opened 2021-07-07
New 'plugin-ready' event and handler

tiberio-baptista:
Adds a new event and event handler to resend `result-selection-ready` event.

## Details
Certain web plugins would get the `result-selection-ready` event from the webapp before initializing their listeners/handlers. 

This PR includes a new webapp event listener called `plugin-ready`, which is handled to resend the same data it would send in the `result-selection-ready` event handler (`ResultsSelected.get()`) to the web plugin.


tiberio-baptista:
This new event is unnecessary after all. The already-in-use `plugin-load` event can be handled in the JSX form file instead, for the same effect.

PR #489, opened 2021-08-10
Protected the hookRemoveRunningTasks, if it happens at same time.

bastiao:
Protected the hookRemoveRunningTasks method, if it happens at same time.

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/taskManager/RunningIndexTasks.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/taskManager/RunningIndexTasks.java
@@ -77,7 +77,7 @@ public class RunningIndexTasks {
 	        while(iterator.hasNext()&& removedTasks<NUMBER_RUNNINGTASKS_TO_CLEAN){
 	            String tId = iterator.next();
                 Task<?> t =  this.taskRunningList.get(tId);
-                if (t.isCancelled() || t.isDone()){
+                if (t != null && (t.isCancelled() || t.isDone())){
                     iterator.remove();
                     removedTasks++;
                 }


dicoogle/src/main/java/pt/ua/dicoogle/taskManager/RunningIndexTasks.java
```diff
@@ -54,7 +55,7 @@ public static RunningIndexTasks getInstance() {
 	}
 
 	public RunningIndexTasks() {
-		taskRunningList = new LinkedHashMap<>();
+		taskRunningList = new ConcurrentHashMap<>();
```
Enet4, line 58: We may need to benchmark this, but there are likely better defaults for these parameters. `16` for the concurrency level is too high. I would try these:

```suggestion
		taskRunningList = new ConcurrentHashMap<>(65_536, 0.75, 4);
```

(increased the initial capacity because resizing the table may be a bit more expensive than with a plain hash map)

dicoogle/src/main/java/pt/ua/dicoogle/taskManager/RunningIndexTasks.java
```diff
@@ -54,13 +56,15 @@ public static RunningIndexTasks getInstance() {
...
+		taskRunningList = new ConcurrentHashMap<>(SOFT_MAX_RUNNINGTASKS, 0.75f, 4);
 	}
 
 	public void addTask(String taskUid, Task<Report> task) {
 		taskRunningList.put(taskUid, task);
-		if (ENABLE_HOOK){
+		if (ENABLE_HOOK && !cleaning){
+		    cleaning = true; // will execute cleaning process
 		    hookRemoveRunningTasks();
+		    cleaning = false; // already cleaned
```
Enet4, line 67: This is definitely not thread safe.

- Changes to `cleaning` are not guaranteed to be visible to other threads (not `volatile`)
- Even if it was `volatile`, the threads could in theory check the condition `!cleaning` at the same time before setting it to `true`, and both enter the cleaning routine.

This can become thread safe by making `cleaning` an `AtomicBoolean` and changing the code to this:
 
```suggestion
		if (ENABLE_HOOK && !cleaning.compareAndSet(false, true)){
		    // will execute cleaning process
		    hookRemoveRunningTasks();
		    cleaning.set(false); // already cleaned
```

Enet4 (changes requested):

dicoogle/src/main/java/pt/ua/dicoogle/taskManager/RunningIndexTasks.java
```diff
@@ -45,22 +47,24 @@
...
+		taskRunningList = new ConcurrentHashMap<>(SOFT_MAX_RUNNINGTASKS, 0.75f, 4);
 	}
 
 	public void addTask(String taskUid, Task<Report> task) {
 		taskRunningList.put(taskUid, task);
-		if (ENABLE_HOOK){
+		if (ENABLE_HOOK && !cleaning.compareAndSet(false, true)){
+		    // will execute cleaning process
 		    hookRemoveRunningTasks();
+            cleaning.set(false); // already cleaned
```
Enet4, line 67: There is mixed whitespace in surrounding lines.
```suggestion
		    cleaning.set(false); // already cleaned
```

This should be fixed in Dicoogle 3, but we might as well keep it consistent here.

bastiao: Dicoogle3: To be honest, not sure if we should keep it consistent. We are just adding here some complexity that in the end we did not take advtange of tracking that list in the way we do. We can discuss it in another scope / hackathon.

Enet4 (approved):

PR #496, opened 2021-08-14
Make the DICOM Storage SCP indexing queue to be asynchronous 

bastiao:
Make the DICOM Storage SCP indexing queue to be asynchronous.
As a consequence, a new task manager instance was created for query to avoid overload of index tasks, and avoid starvation of query tasks.

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
@@ -83,7 +83,7 @@ public class PluginController{
     private final WebUIPluginManager webUI;
     private final DicooglePlatformProxy proxy;
     private TaskManager taskManager = new TaskManager(Integer.parseInt(System.getProperty("dicoogle.taskManager.nThreads", "4")));
-    private TaskManager taskManagerQueries = new TaskManager(Integer.parseInt(System.getProperty("dicoogle.taskManager.nThreads", "4")));
+    private TaskManager taskManagerQueries = new TaskManager(Integer.parseInt(System.getProperty("dicoogle.taskManager.nQueryThreads", "4")));
 
     public PluginController(File pathToPluginDirectory) {
     	logger.info("Creating PluginController Instance");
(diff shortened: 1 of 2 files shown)


dicoogle/src/main/java/pt/ua/dicoogle/server/RSIStorage.java
```diff
@@ -76,8 +76,8 @@
     private BlockingQueue<ImageElement> queue = new PriorityBlockingQueue<>();
     private NetworkApplicationEntity[] naeArr = null;
     private AtomicLong seqNum = new AtomicLong(0L);
+    private static boolean ENABLED_ASYNC_INDEX = Boolean.valueOf(System.getProperty("dicoogle.index.async", "true"));
```
Enet4, line 79: We can just call it `INDEX_ASYNC`.

dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
```diff
@@ -83,7 +83,8 @@ public synchronized static PluginController getInstance() {
     private final WebUIPluginManager webUI;
     private final DicooglePlatformProxy proxy;
     private TaskManager taskManager = new TaskManager(Integer.parseInt(System.getProperty("dicoogle.taskManager.nThreads", "4")));
-    
+    private TaskManager taskManagerQueries = new TaskManager(Integer.parseInt(System.getProperty("dicoogle.taskManager.nThreads", "4")));
```
Enet4, line 86: We should let us specify the thread pool size separately from the other task manager.

bastiao: Yes, forgot to add after our meeting.

bastiao:
I will also propagate this changes to v3.

Enet4 (approved):

Enet4:
As we are aware that critical query requests (e.g. search) go through the task manager, it is wise to reserve a task pool for queries.

Issue #465, opened 2021-05-27
the web plugin does not appear


Kesling515:
I follow the step in this page "https://bioinformatics-ua.github.io/dicoogle-learning-pack/docs/webplugins/" to creat web ui plugin, and  already copy "module.js" and "package.json" above into my "WebPlugins" directory, but the web plugin does not appear.  I try to refresh my page and restart, it is not use. The server log has no information about load it.  
![image](https://user-images.githubusercontent.com/43846775/119800044-3a645280-bf0f-11eb-8826-98e6feec16d9.png)
![image](https://user-images.githubusercontent.com/43846775/119800151-549e3080-bf0f-11eb-852f-a3e01a0c669c.png)

The directory tree should look like this:
 DicoogleDir
 ├── Plugins
 |   └── ...
 ├── WebPlugins
 |   └── aetitle
 |       ├── package.json
 |       └── module.js
 ├── ...
 └── dicoogle.jar

Enet4:
Thank you for reporting. We may need some more information to diagnose the issue.

- Could you share the web plugin's package.json?
- Does the browser report any errors in the console?
- Consider also logging out from Dicoogle (top-right corner) and logging in again, preferably with the default `dicoogle` user or another administrator user if no web plugin appears

Kesling515:
@Enet4
 Hello, thanks for your reply.
1. My plugin's package.json as follows:
 {
  "name": "aetitle",
  "version": "0.1.0",
  "description": "a web ui plugin",
  "main": "module.js",
  "files": [
    "module.js"
  ],
  "scripts": {
    "build": "webpack --config webpack.prod.js",
    "build-debug": "webpack --config webpack.dev.js",
    "build-watch": "webpack --config webpack.dev.js --watch",
    "prepare": "npm run build"
  },
  "author": "Kesling515 <XXXXX@XXX.com>",
  "license": "MIT",
  "repository": {
    "type": "git",
    "url": "Kesling515/aetitle"
  },
  "keywords": [
    "dicoogle",
    "dicoogle-plugin"
  ],
  "dicoogle": {
    "slot-id": "settings",
    "caption": "Aetitle",
    "module-file": "module.js"
  },
  "devDependencies": {
    "@babel/core": "^7.14.3",
    "@babel/preset-env": "^7.14.2",
    "babel-loader": "^8.2.2",
    "babel-minify-webpack-plugin": "^0.3.1",
    "file-loader": "^3.0.1",
[...]

PR #497, opened 2021-08-16
Make webplugins always visible to admin users or if they do not list any roles

Enet4:
Do not hide web UI plugins if their list of roles is empty, even if permissive Web UI is disabled.

It is likely to resolve the main problems reported in #465.


bastiao (changes requested):
If a user is admin, he should have access to all plugins. 
It does not seems to work like this.

bastiao:
LGTM!

PR #503, opened 2021-08-20
Make the DICOM Storage SCP indexing queue to be asynchronous and independent task manager for queries (v3)

bastiao:
This PR introduces two changes:

- Make the DICOM Storage SCP indexing queue to be asynchronous 
- independent task manager for queries (v3)

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/server/DicomStorage.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/server/DicomStorage.java
@@ -78,6 +78,7 @@ public class DicomStorage extends StorageService {
     private BlockingQueue<ImageElement> queue = new PriorityBlockingQueue<>();
     private NetworkApplicationEntity[] naeArr = null;
     private AtomicLong seqNum = new AtomicLong(0L);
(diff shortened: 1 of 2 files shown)


dicoogle/src/main/java/pt/ua/dicoogle/server/DicomStorage.java
```diff
@@ -78,6 +78,7 @@
     private BlockingQueue<ImageElement> queue = new PriorityBlockingQueue<>();
     private NetworkApplicationEntity[] naeArr = null;
     private AtomicLong seqNum = new AtomicLong(0L);
+    private static boolean ASYNC_INDEX = Boolean.valueOf(System.getProperty("dicoogle.index.async", "true"));
```
Enet4, line 81: We can add it later, but the server settings are much nicer to expand in v3, so we can add this setting there as well, leaving the environment variable as an override.

Enet4 (approved):

dicoogle/src/main/java/pt/ua/dicoogle/server/DicomStorage.java
bastiao: We can think about it later.

Enet4: #602

PR #507, opened 2021-08-31
Add missing StudyInstanceUID field in DIMGeneric JSON output

tiberio-baptista:
Added missing `StudyInstanceUID` parameter in `searchDIM` endpoint's output.

## Details
Webplugins of type result-option do not have the `data-study-uid` slot property available, because `StudyInstanceUID` was not being transferred from the REST endpoint.

This PR adds `StudyInstanceUID` to the included mandatory DICOM fields to output in a `searchDIM` request.


Enet4 (approved):
I can validate that his change works just as intended. This was in fact what was already done in Dicoogle 2, but some sort of regression occurred.

PR #514, opened 2021-09-30
Fix the image loader when the thumbnail is a larger image

bastiao:
Fix the image loader when the thumbnail is a larger image.
This happens with in `PopOverImageViewer`.


dicoogle/src/main/resources/webapp/package.json
```diff
@@ -58,7 +58,7 @@
     "react-bootstrap-table": "^4.3.1",
     "react-dom": "^15",
     "react-font-awesome": "https://github.com/tkurki/react-font-awesome/tarball/master",
-    "react-load-image": "^0.1.7",
+    "react-imageloader": "^3.0.0",
```
Enet4, line 61: I had changed this dependency because `react-imageloader` is [no longer maintained](https://www.npmjs.com/package/react-imageloader). Even if we choose to make this reversion now, it may be safer to seek for alternatives. According to my search, `react-image` may fulfill the same criteria and is more popular.

bastiao: Yes, I'm aware about it. The current implementation broke the behaviour/UI in certain conditions. I will try to replace to another more updated component, or I will write it by myself. It is just a few lines of html. :)

Enet4 (approved):
We'll bring this change as is for 3.0.3 and look for alternatives in upcoming versions.

PR #541, opened 2021-12-05
Fix queries using the platform for query plugins that implements QueryInterface

bastiao:
Fix the problem of queries using the platform for query plugins that implements QueryInterface. This will avoid the plugin of query to implement QueryDimInterface, if it is not expected to. 

An example of the issue was: 

```
Task<Iterable<SearchResult>> result = this.platform.query("lucene", "*:*", extraFields);
result.get();
```

It was reported by @rlebre.


Enet4 (approved):
That is right.

rlebre (approved):
Just tested and everything is as supposed! Thank you!

PR #545, opened 2021-12-14
Updated restlet url to https://maven.restlet.talend.com

tiberio-baptista:
The Maven Restlet dependency URLs have become deprecated, and the associated certificates have become invalid since December 11th, making it an insecure connection which fails Maven installations.

This PR replaces the Dicoogle and Dicoogle SDK Restlet dependency URLs to the current URL: `https://maven.restlet.talend.com`


Enet4 (approved):
Yep, this fixed CI as well as some of our local builds, so it's good to merge! Thanks!

PR #547, opened 2021-12-14
Removed not used distribution management endpoints

bastiao:
Removed not used distribution management endpoints 
They were declared already in the parent `pom.xml` and create issues while deploying to remote maven repos.


Enet4 (approved):
The project still builds locally and in CI, so it looks good to me. :+1:

PR #549, opened 2021-12-15
Fix problem with export feature and keyword based query.

bastiao:
Fix problem with export feature and keyword based query.
It can be reproduced with queries such as "Modality:CT" and then export results.
The CSV list will contain only headers, the results of query will be empty.

### DIFF
--- a/dicoogle/src/main/resources/webapp/js/components/search/searchView.jsx
+++ b/dicoogle/src/main/resources/webapp/js/components/search/searchView.jsx
@@ -278,7 +278,8 @@ const Search = createReactClass({
 
     const text = this.state.queryText;
     const provider = this.state.selectedProviders;
-    const params = { text, provider };
+    const keyword = this.isKeyword(text);
+    const params = { text, provider, keyword };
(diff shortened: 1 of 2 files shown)


Enet4 (changes requested):
These changes could probably enter once the proposed changes are made, but this only circumvents the root problem. While the search service in Dicoogle 3 was adjusted to ignore the `keyword` parameter and never perform query expansions unless the `expand` parameter was given, the same was not done for the export service, causing a somewhat severe inconsistency. They definitely should behave the same way for the same query and parameters.

As such, before merging, we need to decide what to do after merging:

- A: Treat the current behavior as a bug, and apply the changes at `ExportCSVToFILEServlet`. The changes are also minor and I have a branch ready. It would become a problem if someone is relying on the current behavior, which I hope not.
- B: Treat the change to the servlet as a breaking change, by only deprecating the `keyword` parameter. This is more expensive and doesn't help fixing expectations much, but satisfies semver compatibility more strictly.

I would rather go with A here, which in that case we do not even need to change the webapp.

PR #568, opened 2022-03-15
Prevent default on export modal button click

### LINKED ISSUE #555: Export to CSV operation refreshes page the first time
Steps to reproduce:

1. Login to Dicoogle
2. Perform a search with some results
3. Click _Export_
4. Define some fields and press either _Save Preset_ or _Export_. The page is refreshed.

Trying again the same steps with the refreshed page usually works.

Enet4:
I believe this fixes #555.


bastiao:
It's seems it fixes!

PR #569, opened 2022-03-15
Update webapp dependencies

Enet4:
In emphasis:

- corejs 3
   - tweaked so that only the needed polyfills are added; reduces js bundle size from 1.1MB to 1.0MB
- Now uses the latest webpack (5)
- remove `react-font-awesome`
   - it was not used in practice
- .babelrc was replaced with babel.config.json
- replaced `babel-eslint` with `@babel/eslint-parser`
- css-loader 6
- ESlint 7
- Required Node.js version for building was bumped to `>=12.22`.


Enet4:
The failure in CI is caused by an unexpected change of behavior from npm v6 to v7. npm/cli#2339 We depend on `npm install` and `npm ci` at `webapp` to populate the `node_modules` folder in `webcore` as well, but the latest versions of npm starting from v7 no longer do this automatically.

For now, I'll add a `postinstall` script to explicitly run `npm install` on `webcore`. It is only slightly redundant in other versions of npm, should still be a pretty fast run.

PR #577, opened 2022-04-16
Change DICOM logging and deprecate XML log

### LINKED ISSUE #528: LogDICOM in Dicoogle 3
We have recent changes to stop emitting log lines to the old DICOMLOG file (#527) in Dicoogle 2. In Dicoogle 3, we need to think of a strategy which will avoid having related issues in the latest versions while fulfilling the necessary DICOM logging needs.

Enet4:
Resolves #528.

- Change implementation to send log line to slf4j logger
   - format is `"[{peer-aetitle}] {type} {additional-info} {params}"`, suggestions on a different structure are welcome
- Disable XML logging by default ("DICOM_Services_Log.xml")
   - can be enabled again with `-Ddicoogle.dicom.xmlLog=true`
- mark related constructs as deprecated, so we can remove them in Dicoogle 4
- tweak `LogLine` class
   - deprecate current constructor, advise users to discard `date`
   - currently still stores `date`, but will be gone in the future

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/DicomLog/LogDICOM.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/DicomLog/LogDICOM.java
@@ -20,23 +20,28 @@
 package pt.ua.dicoogle.DicomLog;
 
 import java.util.ArrayList;
+
+import org.slf4j.Logger;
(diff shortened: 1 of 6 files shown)


bastiao:
That's a good patch. We will move forward with a few tests and get it ready for a new release.

PR #583, opened 2022-06-20
Encode text files and server text responses in UTF-8

### LINKED ISSUE #522: Web services output ISO-8859-1 text in some endpoints
A significant portion of web services in Dicoogle return their text response in a Latin1 encoding, which is not always compatible with the content to be transferred. This happens in those implemented via servlets, such as `/search` and `/searchDIM`.

This is assumed to be the default behavior from Jetty, although UTF-8 is a more reasonable default nowadays.

Enet4:
Resolves #522, making it so that Unicode text values become properly visible in Dicoogle web app and web APIs.

- Override Jetty encoding properties so that responses are encoded in UTF-8.
- Also ensure that some files (tags.xml, roles.xml, CSV exports) are written in UTF-8.

Screenshot dicoogle 2022-06-20 124645

This might only break users expecting CSVs to be in the system's default encoding (or defined via JVM variables). As we are treating #522 as a bug, breaking such uses is an acceptable trade-off in favor of fully supporting Unicode and letting plugin implementers rely on UTF-8 for their web services.

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/core/TagsXML.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/core/TagsXML.java
@@ -24,7 +24,9 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
+import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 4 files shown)


bastiao:
This seems to work pretty well. Also tested with other Swedish templates. 
The CSV also works fine, if JVM is setup with UTF-8, I guess. At least, the exporters seems to work fine on my side.

PR #590, opened 2022-07-04
New/image worker interface

Rui-Jesus:
This PR introduces a new interface to Dicoogle to request directly image services from plugins, instead of using a rest plugin.

This work will serve as a base for an upcoming task to integrate image analysis services in Dicoogle.

Key changes:

-  New image worker interface
- Integration of new interface in plugin manager
- New data structs to manage image rois


Enet4 (changes requested):
Thank you for your contribution. However, we need to put this on hold until the following key points are settled:

- What exactly is being proposed here? Technical issues aside, `ImageWorkerInterface` is insufficiently documented and I cannot infer much from its definition.
- What are the general use cases solved by having this plugin interface? What would be a possible implementation, and does the interface make sense beyond the one presented?
- What added benefits do we have when incorporating this into the SDK? What can the core project do to integrate these plugins and provide value to the archive as a whole?

As this is part of a bigger plan to support image analysis, it may help to disclose what the subsequent steps would be.

Rui-Jesus:
> Thank you for your contribution. However, we need to put this on hold until the following key points are settled:
> 
> * What exactly is being proposed here? Technical issues aside, `ImageWorkerInterface` is insufficiently documented and I cannot infer much from its definition.
> * What are the general use cases solved by having this plugin interface? What would be a possible implementation, and does the interface make sense beyond the one presented?
> * What added benefits do we have when incorporating this into the SDK? What can the core project do to integrate these plugins and provide value to the archive as a whole?
> 
> As this is part of a bigger plan to support image analysis, it may help to disclose what the subsequent steps would be.

Yeah you're out of the loop for this feature, it has been discussed for some with @bastiao.
This is a WIP and improvements are welcome. This PR is not meant to be final, but I put it up because another contribution (private plugin) needs these updates.

For the moment this is needed for the iPATH project. The problem is that we needed a way to request ROIs from WSI to feed image analysis algorithms. A REST plugin would suffice, but another WIP (branch new/machine-learning-provider from BMD fork) has work in progress for the implementation of a new interface for Image Analysis providers. In this contribution there was a need to process images in core (for the moment, only ROI extraction, but more features are expected) and working with REST plugins was inconvenient, so this new interface was created.

This is the main idea:
image

I suggest we discuss better in a meeting.

Rui-Jesus:
This PR is being closed, the work was migrated here: https://github.com/bioinformatics-ua/dicoogle/pull/631

PR #592, opened 2022-07-09
[webapp] Show URI on image meta-data dump 

### LINKED ISSUE #423: Show URI on image meta-data dump
When pressing the "Dump Image" button, one may also be interested in seeing and copying the file's storage URI.

This information is already available in the response payload of a /dump operation, so all that should be needed is to add a place for the URI in the modal.

bastiao:
- Show URI on image meta-data dump. 
- Avoid the text to be cutted if the width of field text is not enough.

Resolves #423.


Enet4 (approved):
Works for me! I had originally pictured a small font label at the top of the table, but this is a bit more consistent and less cluttered, so let's go with this. :+1:

PR #598, opened 2022-07-21
Bulk unindexing for IndexerInterface

### LINKED ISSUE #594: Indexer batched unindexing
The current definition for [`unindex`](https://github.com/bioinformatics-ua/dicoogle/blob/dev/sdk/src/main/java/pt/ua/dicoogle/sdk/IndexerInterface.java#L73) can take a single URI. As most implementations of this imply one index write for each removal, this constitutes a performance bottleneck.

If an overload was provided for unindexing a list of items, implementations could more easily attain better performance by writing all changes to the index in a batch, rather than flushing each operation one by one.

```java
public ? unindex(List<URI> path) throws ?;
```

Enet4:
This proposes an extension to the `IndexerInterface` so that API consumers can request multiple items to be unindexed at once. Resolves #594.

This extension to the API should be carefully reviewed and evaluated to ensure that implementers can use it properly and that it can be well integrated into existing workflows.

Breaking changes are minimal. It is only a problem if an existing plugin already has a method with the exact same signature, which is unlikely. Hence, we are in position to deliver this in 3.4.0, or postpone to version 4 anyway.

---

To ensure some level of quality in this suggestion, we should have:

- [x] at least one proof of concept plugin implementation
   - We could take this as an opportunity to move forward with hosting the open plugins on GitHub. 
- [x] at least one example that the Dicoogle core can consume this method in a usable way without downsides
   - I suggest an extension to the unindexing service so that it accepts multiple URIs

---

Problems resolved in the latest version:

- [x] No way to retrieve the task's unindexing progress from another thread. **Now has a second parameter as a progress callback.**
- [x] Even if the indexer periodically flushes some unindexing operations during the task, other routines cannot work with URIs which have already been unindexed before the task ends completely. This could pose an issue in file removal tasks, because it requires `unindex` to finish before we start cleaning up the files. **The operation is now asynchronous.**

---

Known caveats:
- ~~It is not clear whether we can properly constrain the number of parallel unindexing tasks by pushing this to our indexing task pool. The current indexing task manager depends on task objects returning a different kind of report, which `UnindexReport` does not implement. Maybe it should be subclassed from `Report`?~~ As of the latest version, unindexing tasks can be dispatched by the indexer thread pool.
- Having a second parameter to keep track of progress is quirky. One would naturally expect an `UnindexTask<T>` to extend `Task<T>`, but it is very different from indexing tasks and would make it even harder to attach to the task pool.
- The resulting object is too complex and granular. While there is a lot of power in being able to specify an error for each URI, it is not always the case that files will be unindexed independently, so it may be hard to build such results in batched unindexing.

### DIFF
--- a/sdk/src/main/java/pt/ua/dicoogle/sdk/IndexerInterface.java
+++ b/sdk/src/main/java/pt/ua/dicoogle/sdk/IndexerInterface.java
@@ -25,41 +25,58 @@ import pt.ua.dicoogle.sdk.datastructs.Report;
 import pt.ua.dicoogle.sdk.task.Task;
 
 /**
- * Index Interface Plugin. Indexers analyze documents for performing queries. They may index
- * documents by DICOM metadata for instance, but other document processing procedures may be involved.
+ * Indexing plugin interface.
+ * 
+ * Indexers analyze and record documents for future retrieval.
+ * They are primarily designed to index DICOM meta-data,
+ * which in that case they are accompanied by a query plugin,
+ * and both plugins are called <em>DIM providers</em>.
+ * However, indexers are not restricted to processing DICOM files,
+ * or to retrieving and indexing meta-data.
  *
- * @author Luís A. Bastião Silva <bastiao@ua.pt>
+ * @author Luís A. Bastião Silva <bastiao@bmd-software.com>
  * @author Frederico Valente <fmvalente@ua.pt>
  */
 public interface IndexerInterface extends DicooglePlugin {
 
     /**
-     * Indexes the file path to the database. Indexation procedures are asynchronous, and will return
+     * Indexes the file path to the database. Indexing procedures are asynchronous, and will return
      * immediately after the call. The outcome is a report that can be retrieved from the given task
      * as a future.
      *
      * @param file directory or file to index
-     * @return a representation of the asynchronous indexation task
+     * @return a representation of the asynchronous indexing task
      */
     public Task<Report> index(StorageInputStream file, Object... parameters);
 
     /**
-     * Indexes multiple file paths to the database. Indexation procedures are asynchronous, and will return
+     * Indexes multiple file paths to the database. Indexing procedures are asynchronous, and will return
      * immediately after the call. The outcomes are aggregated into a single report and can be retrieved from
      * the given task as a future.
      *
      * @param files a collection of directories and/or files to index
-     * @return a representation of the asynchronous indexation task
+     * @return a representation of the asynchronous indexing task
      */
     public Task<Report> index(Iterable<StorageInputStream> files, Object... parameters);
 
     /**
-     * Checks whether the file in the given path can be indexed by this indexer. The indexer should verify if
-     * the file holds compatible content (e.g. a DICOM file). If this method returns false, the file will not
-     * be indexed.
-     *
+     * Checks whether the file in the given path can be indexed by this indexer.
+     * 
+     * The method should return <code>false</code> <em>if and only if</em>
+     * it is sure that the file cannot be indexed,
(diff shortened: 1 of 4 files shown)


bastiao:
My opinion is to have it in 3.2.0. 
Thinking aloud: my main concern about the current new API is the not capability to report the URIs unindexed (or perhaps the one it may fail). But I can take a look in our use case and see how it will look like.

Enet4:
> my main concern about the current new API is the not capability to report the URIs unindexed (or perhaps the ones it failed).

That is true. I feared that including that information would overcomplicate the interface, but the cost of not knowing which files were unindexed and which ones were not is possibly too high. I will revise this.

bastiao:
do you have plans to "un-draft" this PR?

Enet4:
> do you have plans to "un-draft" this PR?

I have updated the PR with a plan. It should not become ready for review until there is evidence that this API works well for implementers and consumers. We also have two known problems and further thinking is needed to either resolve them or just accept them as not enough of a problem to merit blocking the extension. Feedback is more than welcome.

Enet4:
Updated the API to include a progress callback. I updated the root message with known caveats.

bastiao:
The unindex bulk API can be beneficial in certain scenarios, and we are currently approaching a milestone where new APIs will be incorporated. It would be appreciated if you could take into account the undraft this pull request, as it has been pending for quite some time. If necessary, we can revisit and iterate on it at a later stage.

Enet4:
> The unindex bulk API can be beneficial in certain scenarios, and we are currently approaching a milestone where new APIs will be incorporated. It would be appreciated if you could take into account the undraft this pull request, as it has been pending for quite some time. If necessary, we can revisit and iterate on it at a later stage.

I have integrated bulk unindexing to the core unindex Web service. It would be better to also have a proof of concept implementation other than the default.

Enet4:
I added this point to the known caveats.

> The resulting object is too complex and granular. While there is a lot of power in being able to specify an error for each URI, it is not always the case that files will be unindexed independently, so it may be hard to build such results in batched unindexing.

After trying to implement this interface, I am convinced that the API needs to be redesigned to be easier to implement.

Enet4:
I have readjusted the Unindex report so that each failure may specify multiple URIs, which is a better reflection of what will happen in most implementations.

sdk/src/main/java/pt/ua/dicoogle/sdk/IndexerInterface.java
```diff
@@ -67,8 +93,63 @@ public default boolean handles(URI path) {
...
+                        }
+                    } else {
+                        // failed to unindex, reason unknown
+                        failures.add(new FailedUnindex(Collections.singleton(uri), null));
+                    }
+                } catch (Exception ex) {
+                    failures.add(new FailedUnindex(Collections.singleton(uri), ex));
+                }
+            }
+            return UnindexReport.withFailures(failures);
```
tiberio-baptista, line 152: If `progressCallback` is null, the UnindexServlet will log “Finished unindexing 0 out of `N` files” for any bulk index, even if there were successful operations.

Is this intended behavior?

Enet4: Well, not really, but I'm not sure if this is a problem with the core components or with the plugins. I updated `UnindexReport` so that the file count is well calculated, but I feel that that log line should not depend on `progressCallback`.

tiberio-baptista (approved):
I have applied the new unindex method for the interface in our index plugins, and it works as expected. Performance is definitely being improved.

LGTM!

bastiao (approved):

PR #606, opened 2022-10-02
Add opt-in `call-shutdown` setting to call plugin shutdown routines on Dicoogle shutdown

### LINKED ISSUE #149: Shutdown plugins when terminating the server

Enet4:
Although I was originally thinking about bringing the shutdown hook to call `shutdown()` on all plugin sets in Dicoogle 4, we felt that there was a need to have this earlier. This PR makes this capability _opt-in_, so that it does not require any breaking changes and does not lead to surprising behaviors.

Resolves #149.

- New server settings XML element `<call-shutdown>` in `<archive>`
   - if true, Dicoogle will call  `shutdown` on (not dead) plugin sets installed
   - this is false by default, but the plan is to make it true by default (opt-out) in Dicoogle 4, and maybe even deprecate it, so that in later versions the behavior becomes mandatory.
   - add test coverage for call-shutdown setting
- add shutdown hook to call  `PluginController.shutdown`
   - Although we've had issues with this when interoperating with Sentry, we currently believe that this functionality is more important than Sentry integration.
- improve `PluginController.shutdown()` by catching and reporting all exceptions individually

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
@@ -278,7 +278,7 @@ public class PluginController {
                 logger.debug("Plugin set {} is shutting down", plugin.getName());
                 plugin.shutdown();
             } catch (Exception ex) {
-                logger.warn("Plugin set {} did not shutdown gracefully", plugin.getName(), ex);
+                logger.error("Plugin set {} did not shutdown gracefully", plugin.getName(), ex);
             }
         }
     }
--- a/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
@@ -263,16 +263,23 @@ public class PluginController {
     /**
      * Stops the plugins and saves the settings
      */
-    public void shutdown() throws IOException {
+    public void shutdown() {
         for (PluginSet plugin : pluginSets) {
-            // TODO: I Think it is better to enable auto-save settings
-            /*Settings settings = plugin.getSettings();
+            // TODO(#605) Consider auto-saving settings on shutdown:
+            /*
+            Settings settings = plugin.getSettings();
             if (settings != null) {
                 settings.save();
             }
             */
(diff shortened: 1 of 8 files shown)


Enet4:
Rebased and updated PR to fix placeholder logic in the shutdown hook.

dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
```diff
@@ -263,16 +263,23 @@ private void initJettyInterface(Collection<PluginSet> plugins) {
...
             */
-            // lets the plugin know we are shutting down
-            plugin.shutdown();
+
+            // let the plugin know we are shutting down
+            try {
+                logger.debug("Plugin set {} is shutting down", plugin.getName());
+                plugin.shutdown();
+            } catch (Exception ex) {
+                logger.warn("Plugin set {} did not shutdown gracefully", plugin.getName(), ex);
```
bastiao, line 281: should we consider this as an error, instead of warning?

Enet4: I am not opposed to that. Change made. :+1:

bastiao (approved):
Great. It seems now that this is also reacting to the `call-shutdown` conf settings.
Tested only with integrated sample plugin (only one drafted here https://github.com/bioinformatics-ua/dicoogle-plugin-sample/pull/14) 
Then also tested integrated with docker stop and it did not went well with shutdown. 
Then did a try with: 
```
docker stop -t 10 dicoogle 
```

and it worked fine. Probably `docker-compose` also add a default time before killing the process, but decided to comment it for awareness.

PR #615, opened 2022-10-15
Added a new reference article to README

bastiao:
Added a new reference article to README.
Also correct @rlebre name in the previous article.


Enet4 (approved):
🎉

PR #637, opened 2023-02-09
DicomStorage: Add guards on unexpected exceptions

Enet4:
Increase robustness and introspection of the DICOM storage service.

- create slf4j logger in `DicomStorage`
- wrap DICOM storage executor around logging executor
- add debug log line at `cstore`
- extend catch to Exception at `onCStoreRQ`
- extend catch to Exception at `Indexer.run`


bastiao (approved):
LGTM.

Tested normal workflow and verify no issue.

PR #638, opened 2023-02-11
[C-MOVE] Fix problem related with infinite loop if skip IS returns 0

bastiao:
Fix problem related with infinite loop if the `skip` method of` InputStream` implementation does not do the job and returns 0.
This can happen with `CipherInputStream` with  `BuferedInputStream` wrapper.

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/server/queryretrieve/DicoogleDcmSend.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/server/queryretrieve/DicoogleDcmSend.java
@@ -655,8 +655,7 @@ public class DicoogleDcmSend extends StorageCommitmentService {
             if (tsuid.equals(info.tsuid)) {
                 try (InputStream fis = info.getInputStream()) {
                     long skip = info.fmiEndPos;
-                    while (skip > 0)
-                        skip -= fis.skip(skip);
+                    skipExactly(fis, skip);
 
                     out.copyFrom(fis);
                 }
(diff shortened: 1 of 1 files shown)


Enet4 (changes requested):
From what I could gather, this change intends to fix an issue where DCM send gets stuck for some implementations of `InputStream`. `CipherInputStream` in particular, is a bit unusual in its implementation. `available` often returns 0 here because according to the [documentation](https://docs.oracle.com/javase/8/docs/api/javax/crypto/CipherInputStream.html), it only returns the number of bytes which have been deciphered, and `skip` will only skip what it can without blocking. Without issuing an actual read, we may indeed get stuck.

On the other hand, forcing us back into reading the whole file into memory would get the archive again in a rough spot when working with very large files. Encryption might not even happen at all in installations, so they could be unnecessarily paying the price in the event of the program receiving false positives from `available` returning 0 anyway.

I was about to accept this with a possibility to opt out of the extra logic, but then I realized that this would only treat the symptom but not the disease. This is happening because `writeTo` is assuming that `skip` can always be used to go through the entire stream. As such, the way I see it, the real root solution is to adjust the way we are skipping an exact number of bytes in the stream.

I have added two commits based on this branch into [bug/c-move-freezes-with-encpstreams-extend](https://github.com/bioinformatics-ua/dicoogle/commits/bug/c-move-freezes-with-encpstreams-extend). One of them makes the eager reading into memory opt-in through a system parameter. The other one replaces the skipping logic in `writeTo`.
Please see that we incorporate this. At the very least, this file read into memory should be opt-out at the process level.

bastiao:
Nice patch and ideas! Indeed, you are right, actually I realize also the root causes while the troubleshooting, and tried different solutions. Just decided to try a less risky solution, but go ahead with your solution. Feel free to incorporate commits here, or close PR and open new one. If it works for you, I will be happy to accept and merge it "now".

Enet4 (approved):
I inserted the two commits and tested the full contribution with two large study c-moves, no issues detected.

PR #643, opened 2023-05-02
Update logging dependencies and add jul-to-slf4j

Enet4:
### Summary

- Update logging dependencies:
   - slf4j to 1.7.36
   - log4j to 2.20.0
- Add `jul-to-slf4j` so that it is possible to collect logs from libraries using java util logging


bastiao (approved):
Let's try :)

PR #644, opened 2023-05-03
imp/Remove uri from default search servlet fields

tiberio-baptista:
This PR removes `uri` from the default search result fields, in order to prevent replicating this information inadvertently.


bastiao:
I would like to understand better impacts of it.

PR #646, opened 2023-05-29
bug/Cast QueryInterface to QueryDimInterface

tiberio-baptista:
This PR ensures that the `getTaskForQueryDim` method explicitly casts its `queryEngine` var into a `QueryDimInterface` instance, instead of `QueryInterface`. The lack of casting to `QueryDimInterface` was causing issues with query providers, namely when trying to use DIM query methods.

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
@@ -605,7 +605,7 @@ public class PluginController {
                     return queryEngine.query(query, parameters);
                 } catch (RuntimeException ex) {
                     logger.warn("Query plugin {} failed unexpectedly", querySource, ex);
-                    return Collections.EMPTY_LIST;
+                    return Collections.emptyList();
                 }
 
             }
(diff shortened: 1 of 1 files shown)


dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
```diff
@@ -618,19 +618,30 @@ public Iterable<SearchResult> call() throws Exception {
...
-        final QueryInterface queryEngine = getQueryProviderByName(querySource, true);
+        final QueryDimInterface queryEngine;
+        QueryDimInterface tmpQueryDimInterface;
+        try {
+            tmpQueryDimInterface = (QueryDimInterface) getQueryProviderByName(querySource, true);
+        } catch (NullPointerException ignored) {
+            tmpQueryDimInterface = null;
+        }
+
+        queryEngine = tmpQueryDimInterface;
```
Enet4, line 629: AFAIK a bad cast would throw `ClassCastException`, not `NullPointerException`. Best to just keep checking dynamically.

```suggestion
        final QueryDimInterface queryEngine;
        QueryInterface tmpQueryDimInterface = getQueryProviderByName(querySource, true);
        if (tmpQueryDimInterface instanceof QueryDimInterface) {
            queryEngine = (QueryDimInterface) tmpQueryDimInterface;
        } else {
            logger.warn("Query plugin {} is not a DIM query provider", querySource);
            queryEngine = null;
        }
```

Enet4 (changes requested):

bastiao:
It's seems ok for me.

PR #649, opened 2023-06-07
Transfer Options - improve web service and UI

### LINKED ISSUE #428: Make Transfer Options page easier to use and understand
The current Transfer Options page leaves a bit to be desired. The "Select All" and "Deselect All" will check and uncheck **all** transfer syntaxes for _**all** SOP classes_. It is a heavy operation which the user might not predict, since the page only presents checkboxes for the SOP class currently selected. Other than this button, transfer syntaxes need to be unchecked one by one, with no extra facilities for disabling or enabling only baseline transfer syntaxes (Implicit VR Little Endian and Explicit VR Little Endian).

Screenshot of the Dicoogle Transfer Options UI

I propose that these aspects are improved:

- ~~There isn't a way to enable a transfer syntax regardless of the SOP class (or at least for those already admitting at least one transfer syntax). This means that allowing a transfer syntax requires looking up every applicable SOP class.~~
   - In retrospective, while there could be interest in this, but it would be hard to get it right: we cannot simply pick all abstract syntaxes because those are likely not to be accepted by intention. It may be best to let users enable a new transfer syntax for each SOP Class they intend to handle.
- The buttons can be disambiguated by adding another set of buttons to enable all transfer syntaxes for the visible SOP class.

Enet4:
### Summary

- Include transfer syntax UID in transfer options response
   - `uid` field is added for each transfer syntax in each SOP class, so that it can be presented in the UI
- Tweak transfer option update endpoint to also accept HTTP methods PUT (as well as POST)
- [webapp] Improve Transfer Options submenu
   - show Transfer Syntax UID in each option
   - simplify and restyle form, reduce text
   - have two "Select all" buttons instead of one
      - One of them changes TSes for the selected SOP class, the other one changes all transfer syntaxes. (resolves #428)

Screenshot 2023-06-07 164939

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/server/SOPList.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/server/SOPList.java
@@ -339,6 +339,7 @@ public class SOPList {
                 JSONObject tsobj = new JSONObject();
                 String name = TransfersStorage.getGlobalTransferMap().get(i);
                 boolean value = ts.getTS()[i];
+                tsobj.put("uid", TransfersStorage.convertTsNameToUID(name));
                 tsobj.put("name", name);
                 tsobj.put("value", value);
 
--- a/dicoogle/src/main/java/pt/ua/dicoogle/server/TransfersStorage.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/server/TransfersStorage.java
@@ -185,6 +185,10 @@ public class TransfersStorage {
         }
         return return_value;
     }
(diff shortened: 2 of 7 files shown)


bastiao:
is it ready to review?

Enet4:
Sure! Please see if the UI changes are suitable and intuitive.

bastiao (approved):
It takes a bit of requests and may provide wrong feedback, as saved appears before requests completes. But it already provide a very good help!

PR #652, opened 2023-06-27
Remove RuntimeException stack trace from query warning log

tiberio-baptista:
This PR adjusts the PluginController to change the warning log when getting a `RuntimeException` from a query plugin (both regular queries or DIM queries).

Instead of writing that the query plugin “failed unexpectedly” and printing the stack trace, the warning now removes the stack trace in favor of indicating the level of the attempted query. 

It can be possible that a query plugin can only handle queries of one or few levels. The plugin controller already deals with this situation by returning an empty list instead of propagating the error, and this empty list can be handled by whatever had requested the query task. Because of these facts, printing the stack trace is a bit overkill, and can flood the Dicoogle logs.

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
@@ -604,7 +604,7 @@ public class PluginController {
                 try {
                     return queryEngine.query(query, parameters);
                 } catch (RuntimeException ex) {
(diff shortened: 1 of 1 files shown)


Enet4:
We may have to put this on hold until we are absolutely sure what behavior we want the core to have. Checking for empty results is not a great way to test for query level support.

Enet4:
We have decided not to move forward with this, and instead require implementers of `QueryDimInterface` to provide implementations more reliably.

PR #655, opened 2023-07-06
Fix #653 - URI can be dumped in meta table

### LINKED ISSUE #653: URI can no longer be found in the meta-dump table
Although we had implemented this in #592, the field `uri` is nowhere to be found anymore.

bastiao:
Fix #653 - URI can be dumped again in meta table


Enet4 (approved):
It works! 🎉

PR #658, opened 2023-07-14
building plugins 

Rajultiwari547:


Enet4:
Sorry, but this does not appear to be a meaningful contribution to Dicoogle. You probably meant to save some of that to your own repository as part of plugin development.

PR #660, opened 2023-07-21
Improve StorageInterface, add method `get` 

Enet4:
This PR proposes a backwards-compatible extension to `StorageInterface` with default method `get`. Unlike `at`, `get` is designed to retrieve up to a single item in storage, so it does not need file system tree traversal or building up iterables and iterators.

A default implementation is provided that uses `at` underneath, but implementors may wish to override it for more efficiency.

This PR also includes a few quality-of-life improvements, including some clarifications to other methods in the plugin interface.

### Summary

- Add default method `StorageInterface#get`, to retrieve a single item from storage
   - update code in core to use it where suitable
   - also correct image servlet to only query for instances to DIM providers if specified
- Fix example code in `at`
- Clarify in `list` that a trailing forward slash means that it's a directory

### DIFF
--- a/sdk/src/main/java/pt/ua/dicoogle/sdk/StorageInterface.java
+++ b/sdk/src/main/java/pt/ua/dicoogle/sdk/StorageInterface.java
@@ -61,7 +61,7 @@ public interface StorageInterface extends DicooglePlugin {
      * This method is particularly nice for use in for-each loops.
      * 
      * The provided scheme is not relevant at this point, but the developer must avoid calling this method
-     * with a path of a different schema.
+     * with a path of a different scheme.
      *
      * <pre>
      * URI uri = URI.create("file://dataset/");
@@ -82,7 +82,7 @@ public interface StorageInterface extends DicooglePlugin {
      *
      * The provided scheme is not relevant at this point,
      * but the developer must avoid calling this method
-     * with a path of a different schema.
+     * with a path of a different scheme.
      *
      * <pre>
      * URI uri = URI.create("file://dataset/CT/001.dcm");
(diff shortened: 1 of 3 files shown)


bastiao (approved):
I understand that it useful. 
I will merge it soon.

Enet4:
@bastiao Since the next Dicoogle version is intended to be a patch version, let's keep this PR on hold until we plan to deploy a feature release.

sdk/src/main/java/pt/ua/dicoogle/sdk/StorageInterface.java
```diff
@@ -62,18 +62,56 @@ public default boolean handles(URI location) {
      * 
      * The provided scheme is not relevant at this point, but the developer must avoid calling this method
      * with a path of a different schema.
```
Enet4, line 64:
```suggestion
     * with a path of a different scheme.
```
```suggestion
     * with a path of a different schema.
```

sdk/src/main/java/pt/ua/dicoogle/sdk/StorageInterface.java
```diff
@@ -62,18 +62,56 @@ public default boolean handles(URI location) {
...
      * @see StorageInputStream
      */
     public Iterable<StorageInputStream> at(URI location, Object... parameters);
 
+    /**
+     * Obtains an item stored at the exact location specified.
+     *
+     * The provided scheme is not relevant at this point,
+     * but the developer must avoid calling this method
+     * with a path of a different schema.
```
Enet4, line 85:
```suggestion
     * with a path of a different scheme.
```
```suggestion
     * with a path of a different schema.
```

Enet4 (commented):
I might as well fix this typo.

tiberio-baptista (approved):
The changes are quite nice, and current behavior of these components remains functional. LGTM!

Enet4:
@bastiao Next version of Dicoogle should be 3.4.0, so we're ready to start merging this and the other ones in the milestone:

- #660
- #598
- #695
- Maybe even #631

Issue #378, opened 2018-11-09
C-FIND response: invalid data for VR 'IS' in SeriesNumber


hkoestin:
Dear all,

I am just struggling with some C-FIND requests on the PACS.
I get back wrong DICOM datasets, having malformed values in e.g. the SeriesNumber tag, wich is of VR IS.

The transmitted data shows the following information for the DIMSE Dataset:
```
(0008,0005) CS (no value available)                             #     0, SpecificCharacterSet
(0008,103e) LO [left intraop 3 ]                                #    16, SeriesDescription
(0020,000d) UI [1.2.826.0.1.3680043.2.594.189609122.591032603.) #    64, StudyInstanceUID
(0020,000e) UI [1.2.826.0.1.3680043.2.594.189609122.591032603.) #    66, SeriesInstanceUID
(0020,0011) IS [1.0]                                            #     4, SeriesNumber
```

**Note: the tag "SeriesNumber" of VR IS with the actual value of "1.0" in it. Why?**

I did cross-check the file with several viewers:
* fo-dicom Dicom.Dump from https://github.com/fo-dicom/fo-dicom/releases/tag/3.0.2
* MicroDicom from http://www.microdicom.com/
* Dicomizer from http://www.roniza.com/products/dicomizer/

All those show the value of "1" in the SeriesNumber tag.

From Dicom.Dump
![image](https://user-images.githubusercontent.com/13849875/48257727-9cbddc00-e413-11e8-9fbf-251b4128630b.png)
From MicroDicom
![image](https://user-images.githubusercontent.com/13849875/48257777-be1ec800-e413-11e8-9763-bcdb2048b3bf.png)
From Dicomizer
![image](https://user-images.githubusercontent.com/13849875/48257817-e4446800-e413-11e8-8d4f-02dc39543cd0.png)

When dumping the file on the web-interface from Dicoogle, I see the following:
![image](https://user-images.githubusercontent.com/13849875/48257881-0fc75280-e414-11e8-88a0-509d7acd35a1.png)

Was it maybe already imported wrongly?

The same file on an Orthanc PACS gives correctly "1" in the value for SeriesNumber when doing a C-FIND. See below:

```
--------------------------------------------------------------------------------
 DIMSE Dataset:
--------------------------------------------------------------------------------
(0008,0005) CS [ISO_IR 100]                                     #    10, SpecificCharacterSet
(0008,103e) LO [left intraop 3]                                 #    14, SeriesDescription
[...]

PR #662, opened 2023-08-10
Fix cast in SeriesNumber mostly affecting C-FIND and C-MOVE services

bastiao:
Just avoid cast issue if the result of SeriesNumber is an integer.

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/server/SearchDicomResult.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/server/SearchDicomResult.java
@@ -279,7 +279,11 @@ public class SearchDicomResult implements Iterator<DicomObject> {
 
                 result.putString(Tag.Modality, VR.CS, (String) sR.get("Modality"));
 
-                result.putString(Tag.SeriesNumber, VR.IS, "" + sR.get("SeriesNumber"));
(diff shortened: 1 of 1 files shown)


dicoogle/src/main/java/pt/ua/dicoogle/server/SearchDicomResult.java
```diff
@@ -279,7 +279,7 @@ public DicomObject next() {
 
                 result.putString(Tag.Modality, VR.CS, (String) sR.get("Modality"));
 
-                result.putString(Tag.SeriesNumber, VR.IS, "" + (String) sR.get("SeriesNumber"));
+                result.putString(Tag.SeriesNumber, VR.IS, "" + sR.get("SeriesNumber"));
```
Enet4, line 282: We might as well add an extra safeguard against #378 regardless of what the underlying index provides.
```suggestion
                String seriesNumber = String.valueOf(sR.get("SeriesNumber"));
                if (seriesNumber.endsWith(".0")) {
                    seriesNumber = seriesNumber.substring(0, seriesNumber.length() - 2);
                }
                result.putString(Tag.SeriesNumber, VR.IS, seriesNumber);
```

bastiao:
Sounds good.

PR #664, opened 2023-08-17
Improve indexer queue worker thread handling

### LINKED ISSUE #663: Indexer thread leaks when disabling DICOM storage
When the DICOM storage service is taken down through the operation to _disable_ DICOM Storage, the indexer thread (which consumes from the queue of stored items and dispatches indexing tasks) is recreated, but the old one stays forever waiting for items which will never arrive, resulting in a thread leak.

Enet4:
This PR resolves some known issues with the DICOM storage indexer worker thread, most importantly:

- The indexer thread is taken down when DICOM storage is stopped (resolves #663).
- If the indexer thread dies, the incident is logged and the DICOM storage service is stopped to prevent further damage. We do expect IT to intervene at this stage. An alternative would be terminating the whole archive.

### Summary

- The indexer thread was given the name `indexer-queue-worker`
- Define indexer thread with a `Runnable` instead of a `Thread` subclass
- Add logic to take down the old indexer worker thread when DICOM storage is stopped.
   - if DICOM storage service is stopped, flip `workerShouldExit` flag and interrupt thread
- Add exception handler to indexer queue worker for damage control. DICOM storage service is disabled automatically, which prevents the archive from accepting files that will never be indexed.

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/server/DicomStorage.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/server/DicomStorage.java
@@ -401,6 +401,12 @@ public class DicomStorage extends StorageService {
     public void start() throws IOException {
         device.startListening(executor);
         indexer.start();
+
+        indexer.setUncaughtExceptionHandler((Thread t, Throwable e) -> {
+            LOG.error("Fatal error in indexer queue worker", e);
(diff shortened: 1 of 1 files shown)


bastiao (approved):
Verified with a few stop/start of DICOM Storage service while keep working with different indexers, and no issues founds.

PR #666, opened 2023-09-08
Provide search results on search-result plugin mount event

tiberio-baptista:
The webapp does not provide the full search results on the `result-selection-ready` event.

This PR fixes this issue by providing both the search results and the search results that were selected via the UI, instead of only providing the selected entries.


Enet4 (approved):
The changes seem solid, and serve a good purpose. As the web UI plugin framework is not currently covered by semver stability guarantees, the upcoming version will be a patch release, 3.3.2. I will leave a warning in the release notes that the API for web UI `result-batch` plugins has changed.

I also filed an [issue at `generator-dicoogle-webplugin`](https://github.com/Enet4/generator-dicoogle-webplugin/issues/17) so that scaffolding webplugins consider this change.

PR #672, opened 2023-10-04
Remove advanced search UI

Enet4:
It was misleading because it hasn't worked for a long time.


bastiao (approved):
Yes, we agreed on that.

PR #673, opened 2023-10-04
Include submitted query and provider in “result-selection-ready” event

tiberio-baptista:
This PR adds the submitted query and provider to the `result-selection-ready` event as a new `request` field.

This further enriched the data that `result-batch` web plugins have access to.

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/server/web/servlets/search/SearchServlet.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/server/web/servlets/search/SearchServlet.java
@@ -228,7 +228,6 @@ public class SearchServlet extends HttpServlet {
                     JSONObject obj = dimModel.getJSONObject(depth, offset, psize);
                     elapsedTime = System.currentTimeMillis() - elapsedTime;
                     obj.put("elapsedTime", elapsedTime);
-                    obj.put("query", query);
                     response.getWriter().write(obj.toString());
                 } catch (Exception e) {
                     logger.warn("Failed to get DIM", e);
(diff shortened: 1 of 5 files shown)


dicoogle/src/main/resources/webapp/js/stores/requestStore.js
```diff
@@ -0,0 +1,28 @@
...
+  get: function() {
+    return {
+      contents: this._contents
+    };
+  }
+});
+
+export { RequestStore };
+
+window.store = RequestStore;
```
Enet4, line 28: Is this line needed?

dicoogle/src/main/java/pt/ua/dicoogle/server/web/servlets/search/SearchServlet.java
```diff
@@ -228,6 +228,7 @@ public void onReceive(Task<Iterable<SearchResult>> e) {}
                     JSONObject obj = dimModel.getJSONObject(depth, offset, psize);
                     elapsedTime = System.currentTimeMillis() - elapsedTime;
                     obj.put("elapsedTime", elapsedTime);
+                    obj.put("query", query);
```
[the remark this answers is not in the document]
tiberio-baptista: Forgot to include in commit. Nicely spotted.

dicoogle/src/main/resources/webapp/js/stores/requestStore.js
tiberio-baptista: Not needed, and neither was the trigger above. Removed both.

bastiao (approved):
Verified with an use case plugin and it works fine.

PR #674, opened 2023-10-06
Webapp maintenance

Enet4:
Updates several dependencies and ensures the presence of `dicoogle-client` 5.2.x in the webapp.

Requires Node.js >=16 to build the webapp (which everyone should have by now anyway).


Enet4:
The current failure with Node.js 21 is caused by a bug in the Dart compiler: https://github.com/dart-lang/sdk/issues/53784

tiberio-baptista:
Please add the `package-lock.json` files to `.gitignore`. These files are generated during deployment based on the related `package.json` files, and are not recessary for the repository.

PR #680, opened 2023-11-06
Fix Web UI Plugin Settings Files Not Handled Properly

tiberio-baptista:
This PR fixes a broken feature of the web UI plugins' config files. Settings files are now properly loaded and provided to the respective web UI plugins. 

## Description
Configurations for a web UI plugin have always been properly handled, so long as the files 
 (which must be named `[WebUIPluginName].json`, see `package.json`'s `name` attribute) are stored in `Plugins/settings`. However, neither the Web UI servlet nor `dicoogle-webcore.js` do any effort to provide or handle the settings so that web UI plugins can access these.

This PR modifies the web UI servlet to include a new `settings` parameter, having the contents of the settings file, should it exist. Furthermore, the webcore was updated to include a new `Settings` field in registered plugins. This allows default webplugin classes to access this field by simply referring to their own attributes (i.e.: `this.Settings`) during the `render(...)` method.

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/server/web/servlets/webui/WebUIServlet.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/server/web/servlets/webui/WebUIServlet.java
@@ -89,7 +89,11 @@ public class WebUIServlet extends HttpServlet {
         List<String> pkgList = new ArrayList<>(plugins.size());
         for (WebUIPlugin plugin : plugins) {
 
-            String pkg = PluginController.getInstance().getWebUIPackageJSON(plugin.getName());
+            JSONObject pkgObject =
+                    JSONObject.fromObject(PluginController.getInstance().getWebUIPackageJSON(plugin.getName()));
+            pkgObject.put("settings", plugin.getSettings());
+
+            String pkg = pkgObject.toString();
             if (pkg == null) {
                 logger.warn("Ignoring web plugin {}", plugin.getName());
                 continue;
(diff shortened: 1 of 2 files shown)


dicoogle/src/main/resources/webapp/js/components/search/result/studyView.js
```diff
@@ -102,7 +102,8 @@ const StudyView = createReactClass({
               uid: item.studyInstanceUID,
               // deprecated data fields
               "data-result-type": "study",
-              "data-result-uid": item.studyInstanceUID
+              "data-result-uid": item.studyInstanceUID,
+              "data-result-patientid": item.patientid
```
Enet4, line 106: With this reasoning, maybe we should also provide more ID fields at the other levels?

dicoogle/src/main/resources/webapp/js/components/search/result/imageView.js
```diff
@@ -142,7 +142,8 @@ const ImageView = createReactClass({
               // deprecated data properties
               "data-result-type": "image",
               "data-result-uri": item.uri,
-              "data-result-uid": item.sopInstanceUID
+              "data-result-uid": item.sopInstanceUID,
+              "data-result-serieuid": item.serieuid
```
Enet4, line 146: Despite the mistakes in the rest of the code, the singular form of _series_ is still _series_.

```suggestion
              "data-result-seriesuid": item.serieuid
```

---

This should not affect any of our WIP plugins.

Enet4 (approved):

PR #683, opened 2024-01-02
Typo fix in CSV export code & docs

tiberio-baptista:
This quick PR fixes a typo present in Dicoogle's REST API pertaining to CSV exporting:

* The endpoint `/export/cvs` was changed to `/export/csv`;
* Related methods/code were refactored;
* The API docs were updated accordingly.


Enet4:
We should not bother changing that endpoint. It is obsolete since Dicoogle version 2.

If I'm not mistaken, you are looking for POST `/exportFile`, available in dicoogle-client as [`issueExport`](https://bioinformatics-ua.github.io/dicoogle-client-js/classes/_index_.dicoogleaccess.html#issueexport), then GET `/exportFile` to download the CSV file.

PR #684, opened 2024-01-26
Change DICOM storage stop behavior to soft stop

Enet4:
#664 introduced a mechanism which stops the DICOM storage service's indexer worker thread when the service is stopped by request. An unwanted side-effect of this is that stopping the storage service will halt any pending requests to index files already in storage, leaving them in the state of stored but not indexed.

### Summary

- let the index queue worker thread finish processing all items in queue before exiting
- do not interrupt ongoing indexing tasks

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/server/DicomStorage.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/server/DicomStorage.java
@@ -80,6 +80,10 @@ public class DicomStorage extends StorageService {
     private BlockingQueue<ImageElement> queue = new PriorityBlockingQueue<>();
     private NetworkApplicationEntity[] naeArr = null;
     private AtomicLong seqNum = new AtomicLong(0L);
+    /**
+     * Whether the index worker thread
+     * should exit once it finishes processing the queue.
+     */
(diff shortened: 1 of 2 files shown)


bastiao (approved):
LGTM

Enet4:
Putting this on hold because an index worker thread already waiting on `take` will not be awoken when the service is stopped.

Enet4:
Mostly replaced `take` by `poll`, which can take a timeout. It might introduce some teeny trace work while waiting for items to index, but it's much better than leaking the thread.

There is also room for implementing a hard stop in the future, should we ever want that.

bastiao:
after discussion, we decided to reuse the thread.

tiberio-baptista:
I can say that this is working as intended now. Indexing continues to work for a TCP connection, even if the storage service is stopped midway.

Subsequent TCP connections are not possible, as expected.

tiberio-baptista (approved):
The PR works as intended.

I ran a traffic generator that sends DICOM files over CSTORE, and studies are no longer interrupted if they are mid-transfer. Subsequent transfers cannot be performed, as intended. 

LGTM!

PR #686, opened 2024-03-14
[core] Add kill switch for dead plugin sets

Enet4:
This implements a safety measure against misconfigured or malfunctioning plugins. If the JVM property `dicoogle.deadPluginKillSwitch` is set to `true` (case insensitive), any dead plugin will kill the Dicoogle process, thus requiring devops to investigate and fix the issues reported in the logs.


bastiao:
may be we can set as default behavior ?

Enet4:
> may be we can set as default behavior ?

Can be done. I am just not sure how disruptive this change might be, since it could break an existing installation because of a plugin that we might not even care about. I say we test internally and draw our own conclusions.

PR #690, opened 2024-05-23
Webapp maintenance (2024)

Enet4:
### Summary

- dependency: ~~react-router 4~~
   - Reverted, to be tried again at a later date.
- dev-dependency: css-loader 7
- dev-dependency: sass-loader 14
- dev-dependency: style-loader 4
- dev-dependency: webpack-dev-server 5
- dev-dependency: eslint 9
- `npm update` so that other dependencies are refreshed
- `npm update` on webcore
- Require a minimum of Node.js 18 to build the webapp


tiberio-baptista (approved):
Webapp continues to work as intended, with these new changes.

Looks good for me!

dicoogle/src/main/resources/webapp/js/external-requires.js
```diff
@@ -5,7 +5,8 @@
  */
 const EXTERNAL_MODULES = {
   react: require("react"),
-  "react-router": require("react-router"),
+  "react-router": require("react-router-dom"),
```
Enet4, line 8: Considering that this is a breaking change to any web plugin using react-router, I might as well drop the external module instead.

```suggestion
```

Enet4:
We've been testing the webapp with these changes for a while. For the most part things seem to be in good order.
However, we currently depend on `react-router` in one of our web plugins so that it can navigate back to the search page. With the updates of react-router, `hashHistory` is no longer available. We may need to identify a migration path, making changes here if necessary.

bastiao:
> We've been testing the webapp with these changes for a while. For the most part things seem to be in good order. However, we currently depend on `react-router` in one of our web plugins so that it can navigate back to the search page. With the updates of react-router, `hashHistory` is no longer available. We may need to identify a migration path, making changes here if necessary.

This will be on-hold until this situation is solved.

Enet4:
I have decided to revert the upgrade of react-router so that it can be worked on separately. When doing so, we must not forget to test for routing capabilities within a web UI plugin.

I'm thinking that at the same time we should try a better React integration again, one which does not force us to create multiple rendering contexts.

bastiao (approved):

PR #691, opened 2024-05-23
Clean up logging around indexing tasks

Enet4:
### Summary

- lower level of some log lines which are not very useful (info to debug)
- remove commented out code

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
@@ -454,7 +454,6 @@ public class PluginController {
         for (QueryInterface p : plugins) {
             names.add(p.getName());
         }
-        // logger.info("Query Providers: "+Arrays.toString(names.toArray()) );
         return names;
     }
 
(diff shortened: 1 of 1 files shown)


dicoogle/src/main/java/pt/ua/dicoogle/plugins/PluginController.java
```diff
@@ -790,7 +776,7 @@ public void remove(URI uri) {
     }
 
     public void doRemove(URI uri, StorageInterface si) {
-        if (si.handles(uri)) {
+        if (Objects.equals(si.getScheme(), uri.getScheme())) {
```
bastiao, line 779: I don't agree with this change. By design, the `handles` method should be responsible for this.

Enet4: The method `StorageInterface#handles` is deprecated, as proposed and established in #257 seven years ago. All I did was replace it with the canonical implementation, which must always be a scheme comparison.

Leaving the method in the interface was particularly error prone before Dicoogle 3, which required implementers to replicate this in all plugins. Since Dicoogle 3, it still does not stop implementers from replacing it with something else, making it unnecessarily error prone.

bastiao: handles should be in place. For instance in case we want to keep blob:// or blob+ssl://.

Enet4: I have reverted this change for now, so that discussion can be continued separately.

bastiao (approved):

PR #694, opened 2024-06-27
Add robustness to DIMGeneric against BodyPartThickness

Enet4:
This fixes a situation found in the wild, where the matching of certain search results via C-FIND would lead to a server error.

### Summary

At `DIMGeneric#fillDim`, do not assume that `BodyPartThickness` is a string, because some index sources might save it as a double.


tiberio-baptista (approved):
LGTM!

PR #700, opened 2024-09-13
Tweak the way unindex task starts are logged

Enet4:
Only log after resolving indexer and indicate the name, thus clearing any eventual confusion of this line appearing multiple times in repeat.


tiberio-baptista (approved):
Tested locally.

The line only appears once, as suggested.

LGTM!

PR #702, opened 2024-10-24
Optimize task management page

Enet4:
### Summary

- Restrain task status view to showing the 200 most recent tasks, so that the page can still open fine on long-running Dicoogle instances without saturating the browser.
- Show how many tasks are presented and the number of tasks not shown in the header above the task list.
- Update task status style so that each task is more compact.

See screenshot.

Screenshot 2024-10-24 095303


tiberio-baptista (approved):
Tested this by running locally and spamming index tasks until the UI displayed 200 tasks.

The new visual is better and less bloated, and I've found with 200 tasks on-screen. Logs show no issues as well. LGTM!

PR #704, opened 2024-11-01
Fix #703 - Fix error when providers are null

### LINKED ISSUE #703: How to display thumbnail images
Please help me, there is currently a DICOM file that does not display thumbnails in this project, but can be viewed using DICOM VIEWER（https://milung.github.io/dicom-editor/dist/）. 
I am currently using the default configuration, how can I make it display thumbnails?

Here is my DICOM file
[1.2.392.200036.9125.0.19950720112207.zip](https://github.com/user-attachments/files/17595370/1.2.392.200036.9125.0.19950720112207.zip)

Whiteyingshan:
Fix #703 - Fix error when providers are null


Enet4 (approved):
Looks OK. It probably would have been better if this list could never be null, but such a change is only likely to happen in a major revision bump now.

PR #705, opened 2024-11-13
Make ControlServices comply with autostart server settings

Enet4:
- do not start the service(s) if their respective autostart setting is set to false, a regression probably existing since Dicoogle 3.0.0
- remove unused variable `webServicesRunning` in `ControlServices`
- remove use of raw types in `startStorage()`
- remove outdated TODO note
- refactor `ControlServices#startStorage()` to return whether the service started (`true`) or was already running (`false`) and raise an exception on error
- adjust `ServicesServlet` according to changes to `ControlServices`, and for the output to be more precise and informative

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/server/ControlServices.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/server/ControlServices.java
@@ -40,14 +40,13 @@ public class ControlServices {
     private static ControlServices instance = null;
     // Services vars
     private DicomStorage storage = null;
-    private boolean webServicesRunning = false;
     private boolean webServerRunning = false;
     private QueryRetrieve retrieve = null;
 
     private DicoogleWeb webServices;
 
     private ControlServices() {
-        startInicialServices();
+        autoStartServices();
     }
 
     public static synchronized ControlServices getInstance() {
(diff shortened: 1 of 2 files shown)


tiberio-baptista (changes requested):
Autostart is now being respected, and I've confirmed this with a local run.

I've confirmed that `ControlServices#startStorage()` behaves as suggested in the PR.

I did find an issue with `ServicesServlet`: though I'm not sure that this was present before, I noticed that `GET /management/plugins/` returns the following response body:

```json
{
  "isRunning": false,
  "port": -1,
  "autostart": false
}
```

Please confirm if this is expected. Otherwise, the PR is good and I can approve.

Enet4:
> I did find an issue with `ServicesServlet`: though I'm not sure that this was present before, I noticed that `GET /management/plugins/` returns the following response body:
> 
> ```json
> {
>   "isRunning": false,
>   "port": -1,
>   "autostart": false
> }
> ```
> 
> Please confirm if this is expected. Otherwise, the PR is good and I can approve.
> 
> Personally, I would expect this endpoint to list the available plugins.

That is a curious finding. That endpoint is not part of the API, but it was probably captured by the service status servlet anyway, which tried to fetch the status of a service which does not exist.

This can be improved, but it is hardly problematic, nor does it not look like a regression from this PR.

tiberio-baptista:
> This can be improved, but it is hardly problematic, nor does it not look like a regression from this PR.

For clarity's sake, I only found and tested that endpoint because I went into the code to see what endpoints that servlet was mapped to, and that one was included.

I have created a new issue to track this, though of course it is of low priority. With nothing else to add, I'll be approving this PR.

PR #709, opened 2024-12-16
Remove dangling /management/plugins/ endpoint

### LINKED ISSUE #708: GET to /management/plugins returns nothing useful
> I did find an issue with `ServicesServlet`: though I'm not sure that this was present before, I noticed that `GET /management/plugins/` returns the following response body:
> 
> ```json
> {
>   "isRunning": false,
>   "port": -1,
>   "autostart": false
> }
> ```
> 
> Please confirm if this is expected. Otherwise, the PR is good and I can approve.
> 
> Personally, I would expect this endpoint to list the available plugins.

That is a curious finding. That endpoint is not part of the API, but it was probably captured by the service status servlet anyway, which tried to fetch the status of a service which does not exist.

This can be improved, but it is hardly problematic, nor does it not look like a regression from this PR.

_Originally posted by @Enet4 in https://github.com/bioinformatics-ua/dicoogle/issues/705#issuecomment-2539749272_

Enet4:
Resolves #708.

### Summary

- convert int constants for service type to an enum
- remove servlet handler for "/management/plugins/"
- document servlet operation methods


tiberio-baptista (approved):
LGTM!

PR #722, opened 2025-04-23
Add data store failure detection in DICOM storage service

### LINKED ISSUE #721: Storage SCP returns status code 00H on C-STORE-RSP when file store fails
If the storage provider fails to store a file, the DICOM storage service will not treat this as an error and continue normally, eventually responding to the SCU with status code 0.

This happens because the method `StorageInterface#store(DicomObject,Object...)` does not have `throws IOException`, which would be used to raise errors in the storage process, and also because the same method returning `null` is treated as something to ignore rather than something to warn the SCU about.

The consequence is that the SCU will not find anything wrong with the transfer even though it was clear from the server that the store could not proceed. In reality, the client device can only be sure that the data is safe through storage commitment, but Dicoogle does not provide storage commitment either.

Enet4:
Should resolve #721.

Caveats:
- it's an SDK breaking change
- reports success whenever at least one storage provider succeeds, meaning that others could have failed when there is more than one. The scenario of having multiple storage providers at once is not that well supported anyway.

### Summary

- [sdk] Change signature of `StorageInterface#store(DicomObject,Object...)` so it `throws IOException`
- [sdk] Adjust documentation of `StorageInterface` a bit
- Detect data storage failure in `DicomStorage`

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/server/DicomStorage.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/server/DicomStorage.java
@@ -278,13 +278,30 @@ public class DicomStorage extends StorageService {
 
             Iterable<StorageInterface> plugins = PluginController.getInstance().getStoragePlugins(true);
 
+            int success = 0;
+            int failures = 0;
             for (StorageInterface storage : plugins) {
-                URI uri = storage.store(d);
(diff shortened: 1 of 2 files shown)


Enet4:
Dropping this in favor of #750.

PR #725, opened 2025-06-17
Fix logout and root page loading dynamics

Enet4:
### Summary

- [webapp] Fix runtime error when trying to log out
   - It was a mistake in the retrieval of the router history object
- [webapp] Fix nuances of loading of root page when the user navigates to `/#/` directly
   - Remove the loading page as an index route
   - Route to login page only after fetching from user store (even when there is no user session on local record)


tiberio-baptista (approved):
Tested locally. Login and logout work without any errors.

I did notice that the following warning is shown in the login form, but this can be adjusted in another PR:

image


Otherwise, LGTM!

PR #736, opened 2025-07-30
[sdk] Mark ML provider API as experimental

Enet4:
Hopefully this is good enough to send the message.


bastiao (approved):

bastiao:
If there are issues in the YAML, I’ll fix them later, as it's been stuck for a while and the build actions still need to be implemented.

PR #738, opened 2025-08-04
Update dicoogle-client to v6

Enet4:
This allows us to check whether we are ready to start using `dicoogle-client` v6 everywhere.

### Summary

- dicoogle-client 6.0.0
- update servicesStore to use promise-based operations
- update exportStore
- Update Web UI module servlet to support ES module dicoogle-client


tiberio-baptista (approved):
I have ran these changes locally.

I have tested both the webapp and webplugins after rebuilding, and I have not found any issue while experimenting with the webapp and plugins.

LGTM!

PR #739, opened 2025-08-22
DICOM storage: save sending and receiving AE titles in FMI of stored objects

Enet4:
This will fill in the standard file meta information group attributes relative to the sending and receiving application entities, so that some level of provenance can be inferred from it.

This can be disabled by setting the server setting `storage.save-ae-titles` to false.

Since dcm4che2 does not recognize these attributes, we also needed to include them in the Dicoogle index `TagsStruct`.

#### Pending questions

- I added _Sending Application Entity Title_ and _Receiving Application Entity Title_. Should I also add _Sending Presentation Address_ and _Receiving Presentation Address_? https://dicom.nema.org/medical/dicom/current/output/chtml/part10/chapter_7.html#sect_7.1.1.1 
   - We have decided not to include it for now.


bastiao:
Should be undraft?

PR #740, opened 2025-10-09
Provide baseline access control to web services

### LINKED ISSUE #98: [webapp] allow role-based access control to features
Implement an hierarchy of user roles such as some features are reserved for higher roles. For instance, index/unindex operations should only be performed by certified administrators.

Enet4:
Should resolve #98 for Dicoogle core services. Security of services in extensions still needs to be upheld by each plugin.

### Summary

- Add another mechanism to stay logged in using a cookie
   - `DICOOGLE_SESSION` is automatically provided on login and is looked up when the `Authorization` header is not found.
- Require to be logged in for most core Dicoogle services
   - Web service extensions via plugins are not included here
   - Require user logged in to be an admin in management related operations
- Add server setting "allow-unauthorized"
   - If `true`, access control is disabled like in dicoogle<=3.5.0
   - A warning is displayed when starting the web server with this flag enabled
- Extend Dicoogle platform interface so that plugins can know that a user is authenticated

### Pending issues

- [x] Requests to `webui/module` and `/dic2png` (for the thumbnails) in the webapp do not have the authentication token at the moment, so they will always fail.
   - Resolved by setting a session cookie.
- [x] Errors in user presets, likely a regression in another PR.
   - Fixed in #738

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/core/settings/LegacyServerSettings.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/core/settings/LegacyServerSettings.java
@@ -430,6 +430,16 @@ public class LegacyServerSettings implements ServerSettings {
             this.accessControlAllowOrigins = origins;
         }
 
+        @Override
+        public void setAllowUnauthorized(boolean allowUnauthorized) {
+            // no-op, not supported by legacy settings
+        }
+
+        @Override
+        public boolean isAllowUnauthorized() {
+            return true;
+        }
+
         @Deprecated
         public void setWebServices(boolean webServices) {
             this.webServices = webServices;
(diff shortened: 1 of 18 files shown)


tiberio-baptista (approved):
Tested locally.

After a few issues with local maven dependencies and cache, I was able to validate the following behaviors:

* Usage of web UI:
  * Log in and be able to use management features with an admin account
  * Log out in other tab, confirm that former tab becomes unable to do anything until new login session is issued
* Usage of web API
  * Confirmed authentication is required for managerial Dicoogle API requests (enable/disable storage, list/create users)
  * Confirmed that valid DICOOGLE_SESSION cookie allows API usaeg if `Authorization` header is absent
* Confirm `allow-unauthorized` disables these auth features, and warning is issued in Dicoogle logs.

I believe everything is in good order, so I'm approving this PR.

PR #741, opened 2025-10-11
Offer custom welcome messages at the login page

Enet4:
This provides a nice-to-have: updated, customizable messages for when a user is about to log in to a Dicoogle instance.

Screenshot_2025-10-11_16-30-17

Welcome messages can be described as simple HTML text, each message separated by two consecutive newlines (`\n\n`). If a custom welcome file is not provided, Dicoogle will use an embedded listing which you should take a look in the new file `welcome.html`. Feedback on other messages to provide are welcome.

### Summary

- [server] add `/welcome` servlet
   - returns one of the listed messages at random, content type `text/plain`
   - loads messages from resource file `welcome.html` or from a custom file (`-Ddicoogle.welcomeFile`)
- [webapp] add `domify` dependency to introduce an HTML sanitization step
- [webapp] update login view to show a welcome message from the ones listed

### DIFF
--- a/dicoogle/src/main/java/pt/ua/dicoogle/server/web/servlets/WelcomeServlet.java
+++ b/dicoogle/src/main/java/pt/ua/dicoogle/server/web/servlets/WelcomeServlet.java
@@ -86,7 +86,7 @@ public class WelcomeServlet extends HttpServlet {
             if (!msg.isEmpty()) {
                 messages.add(msg);
             }
-            if (!messages.isEmpty()) {
+            if (messages.isEmpty()) {
                 messages.add(FALLBACK_MESSAGE);
             }
 


dicoogle/src/main/java/pt/ua/dicoogle/server/web/servlets/WelcomeServlet.java
```diff
@@ -0,0 +1,107 @@
...
+                } else {
+                    msg += line + "\n";
+                }
+            }
+            if (!msg.isEmpty()) {
+                messages.add(msg);
+            }
+            if (!messages.isEmpty()) {
+                messages.add(FALLBACK_MESSAGE);
+            }
```
Enet4: Oopsie. This adds the fallback welcome message when the list is not empty, instead of when it is.

```suggestion
            if (messages.isEmpty()) {
                messages.add(FALLBACK_MESSAGE);
            }
```

PR #748, opened 2025-12-17
Adjust WebUIPluginManager and WebUIPluginTest for test reliability

Enet4:
We have encountered an odd scenario where the JDK would try to retrieve the web UI plugin directory incorrectly in the tests, resulting in false positives. This form should be more resilient.

### Summary

- Replace `assert`s with hard checks which throw exceptions in `WebUIPluginManager#load`
- Tweak test1 dir resource retrieval in `WebUIPluginTest#loadTest1`, adding a fallback to webui plugin path resolution


bastiao (approved):
it works

PR #752, opened 2026-07-15
Extend plugin set dead condition to failures on platform injection

Enet4:
If one of the plugins or the entire plugin set throws an exception during platform injection, mark it as dead, or shut down Dicoogle entirely if the kill-switch is enabled.

This makes serious problems with the plugins less likely to be overlooked.


tiberio-baptista (approved):
Tested this in CI by replacing a specific plugin of our's XML conf with non-XML junk.

A raised `NullPointerException` is now causing the kill switch to trigger and shut down the process. It was previously making the Dicoogle process hang during setup.

LGTM!

bastiao:
@Enet4 perhaps this can be merged?

Enet4:
> @Enet4 perhaps this can be merged?

Sure.

